PR #47, opened 2018-11-15
[update] Fires events when the visibility of an element changes

sabieber:
The details of the event include the element itself and the new visibility of that element.

This brings back an event present in the legacy ui for reacting to UI changes (i.e. opening and closing the sidebar)

Issue was reported in the google group: https://groups.google.com/forum/#!topic/pdfnet-webviewer/wt20LMXOSpU

### DIFF
--- a/src/redux/actions/exposedActions.js
+++ b/src/redux/actions/exposedActions.js
@@ -1,5 +1,5 @@
 import isDataElementPanel from 'helpers/isDataElementPanel';
-
+import { fireEvent } from 'helpers/loadDocument';
 // viewer
 export const enableAllElements = () => ({ type: 'ENABLE_ALL_ELEMENTS', payload: {} });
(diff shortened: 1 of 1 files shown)


sabieber:
Thanks for the review and your feedback, highly appreciated!
I updated the pull request to include your suggested changes and also tested it with different `openElement`, `closeElement` and `setActiveLeftPanel` calls, it seems to work fine now.

Let me know if there is something else I should change.

justinjung04:
Thanks for the updates!
I also noticed that when opening left panel for the first time, visibilityChanged events gets triggered for `leftPanel`, but not for `notesPanel`. I believe we should trigger an event for the active panel too when we open or close `leftPanel`.

sabieber:
Thanks again for the feedback. I included your suggestion in the latest commit

PR #48, opened 2018-11-16
[update] Allows for custom annotation date format

sabieber:
This adds the public API function `setNoteDateFormat` for customizing the date format string used for the creation date of annotations and replies.

### Motivation
This is especially useful for customers of different locales as its possible to choose an appropriate format string.

### Test
`readerControl.setNoteDateFormat('DD.MM.YYYY HH:MM')`

### Discussion
Please let me know if the name of the API should be altered or if there is a better way to add such API.

Thanks for reviewing this PR :)


justinjung04:
This is a great API! Could you actually add the default value to the initial state?

PR #54, opened 2018-11-20
[update] Resolves #49 by adding a function for rendering custom thumbnail content

### LINKED ISSUE #49: API for customizing thumbnails in thumbnail panel
As already discussed in the google group ([topic](https://groups.google.com/forum/#!topic/pdfnet-webviewer/12Gwl4l4kwg)) we need an API for customizing the rendered content of the `Thumbnail`s in the `ThumbnailsPanel`.

In my opinion it would be best for everyone if this was a generic API that could be used by a lot of users so I would like to implement such API.

Would there be requirements that I need to match so the proposed API would be merged by you?

My first idea would be a setter for providing an `activator`/`customizor` function that is called when a Thumbnail is rendered to alter the DOM contents. Does this sound like a good idea for general use?

What we really need is to display two small buttons next to the page number on the Thumbnail (one for download and one for another feature of our customized viewer).

sabieber:
This adds the public API function `setThumbnailCustomContentRender` for adding custom DOM elements to the thumbnail container during rendering.

### Motivation
We need to be able to add custom elements (in pur case individual download buttons) next to the thumbnails in the panel. As this could be a usecase for more users we would like to introduce this API (and because we want to maintain upgradability by not altering the React sources ourselves).

### Test
```
readerControl.setThumbnailCustomContentRenderer(function (index, react) {
      return react.createElement('p', null, 'Custom page: ' + index);
});
```

### Discussion
Please let me know if the name of the API should be altered or if there is a better way to add such API.

Thanks for reviewing this PR :)

### DIFF
--- a/src/apis/index.js
+++ b/src/apis/index.js
@@ -61,6 +61,7 @@ import disableTools from './disableTools';
 import isToolDisabled from './isToolDisabled';
 import setNotesPanelSort from './setNotesPanelSort';
 import updateOutlines from './updateOutlines';
+import setThumbnailCustomContentRenderer from './setThumbnailCustomContentRenderer';
 
 export default {
   loadDocument,
@@ -125,5 +126,6 @@ export default {
   disableTools,
   isToolDisabled,
   setNotesPanelSort,
-  updateOutlines
+  updateOutlines,
+  setThumbnailCustomContentRenderer
 };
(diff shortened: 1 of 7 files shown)


justinjung04:
Closing as per discussion in #49.

PR #65, opened 2018-12-04
Pull request to fix PageNav display

awy3:
The page nav would go from "(1/3)" to "1 (1/3)" when switching between files. 

image

This pull request is to fix that from happening.


awy3:
Need to check something else. Problem might be due to some other issue

awy3:
This was somewhat working as intended (it's using the incorrect format, but the formatting is fine)

PR #68, opened 2018-12-06
[update] Resolves #62 by having annotation popup reappear on mouse up

### LINKED ISSUE #62: Possible improvement to Annotation Popup behavior
On current Demo Viewer -  

1. Create Rectangle Annotation
2. Select the same annotation
3. Click again on the annotation (in between the white circle indicators on the annot).
4. Annotation Popup disappears

Popup only reappears if you click on the white indicators while still selected.

Possible Desired/expected functionality - 
Clicking anywhere on the selected annotation pops up the menu. Popup only goes away if one of it's button's are clicked.  

Users had a very hard time getting back the popup until they were able to figure out clicking on the white indicators got it back.  We had a few complain about it until we explained the usage.

--------
Also, possibly separate issue -
You are unable to recover back to the popup form the style menu.  The only way to get out of the popup style menu is to click off the menu.  A small back or X button somewhere would allow the user to backtrack and keep the annotation selected.

awy3:
Updated the 'componentDidUpdate' of the AnnotationPopup to have it open on mouse up when needed. Before it only open when a new annotation was selected.

### DIFF
--- a/src/components/AnnotationPopup/AnnotationPopup.js
+++ b/src/components/AnnotationPopup/AnnotationPopup.js
@@ -97,6 +97,7 @@ class AnnotationPopup extends React.PureComponent {
         canModify: core.canModify(annotation)
       });
     } else {
+      //this.props.closeElement('annotationPopup');
       this.setState({ ...this.initialState });
     }
   }
--- a/src/components/AnnotationPopup/AnnotationPopup.js
+++ b/src/components/AnnotationPopup/AnnotationPopup.js
@@ -97,7 +97,7 @@ class AnnotationPopup extends React.PureComponent {
         canModify: core.canModify(annotation)
       });
     } else {
-      //this.props.closeElement('annotationPopup');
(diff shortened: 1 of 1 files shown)


src/components/AnnotationPopup/AnnotationPopup.js
```diff
@@ -81,7 +97,6 @@ class AnnotationPopup extends React.PureComponent {
         canModify: core.canModify(annotation)
       });
     } else {
-      this.props.closeElement('annotationPopup');
```
justinjung04: We actually need this line, otherwise the annotation popup remains open when the annotation is deselected from the notes panel

awy3: I put back the line of code that was removed. 

https://github.com/PDFTron/webviewer-ui/pull/68/commits/8bb854e8129f588ce6bdf58912bc5b899cd41175

PR #112, opened 2019-01-02
[bugfix] fixes #111

### LINKED ISSUE #111: File extension check should ignore URL query string
When creating a PDFTron.WebViewer instance, if pdftronServer is set and initialDoc points to a URL with query string parameters (cache busting, etc) the browser dev tools get spammed with an error about an invalid file extension.  

The file extension check should be chopping off the URL query string before trying to determine the file extension.

ZhijieZhang:


justinjung04:
Fixed by 5a154cfff2607c4aa48b1f66afafd4775d9386a4

PR #114, opened 2019-01-03
Add dutch translation

EdwinKorsten:
This is a translation file for the Dutch language, so that dutch readers can use PDFTron in Dutch.


EdwinKorsten:
Thank you for the quick approval. That motivates to make more pull-requests.

Issue #128, opened 2019-01-14
API for changing which colour palette to show by default


ZhijieZhang:
An annotation may have text colour, stroke(border) colour and fill colour. Right now the default colour palette is determined by the following code:
```javascript
const {  TextColor, StrokeColor, FillColor } = this.props.style;

return { 
  colorPalette: TextColor ? 'text' : StrokeColor ? 'border' : FillColor ? 'fill' : ''
};
```
If an annotation has text colour and stroke colour, the default palette will always be text colour palette. However, some customers would like to show a different(by their choice) palette when an annotation has more than one colour property.

ZhijieZhang:
This issue has been resolved in: https://github.com/PDFTron/webviewer-ui/commit/712d81f9ec1d3a702462011d2e88122fc0fa1b88

PR #136, opened 2019-01-18
[update] setDefaultColorPalette API Implemented

JSpark22:
setDefaultColorPalette API fixes issue #128.

### DIFF
--- a/src/components/AnnotationPopup/AnnotationPopup.js
+++ b/src/components/AnnotationPopup/AnnotationPopup.js
@@ -89,6 +89,7 @@ class AnnotationPopup extends React.PureComponent {
   onAnnotationSelected = (e, annotations, action) => {
     if (action === 'selected' && annotations.length === 1) {
       const annotation = annotations[0];
+      console.log(annotation.Subject);
       this.setState({
         annotation,
         canModify: core.canModify(annotation)
@@ -183,6 +184,7 @@ class AnnotationPopup extends React.PureComponent {
 }
 
 const mapStateToProps = state => ({
+  test: selectors.getActiveToolName(state),
   isNotesPanelDisabled: selectors.isElementDisabled(state, 'notesPanel'),
   isDisabled: selectors.isElementDisabled(state, 'annotationPopup'),
   isAnnotationStylePopupDisabled: selectors.isElementDisabled(state, 'annotationStylePopup'),
--- a/src/components/AnnotationStylePopup/AnnotationStylePopup.js
+++ b/src/components/AnnotationStylePopup/AnnotationStylePopup.js
@@ -51,6 +51,7 @@ class AnnotationStylePopup extends React.Component {
     return(
       <div className={className} data-element="annotationStylePopup" onClick={() => closeElement('annotationPopup')}>
         <StylePopup 
+          annotationType = {annotation.Subject}
           style={style}
           isFreeText={isFreeText}
           onStyleChange={this.handleStyleChange}
--- a/src/components/StylePopup/StylePopup.js
+++ b/src/components/StylePopup/StylePopup.js
@@ -1,10 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 
 import ColorPaletteHeader from 'components/ColorPaletteHeader';
 import ColorPalette from 'components/ColorPalette';
 import Slider from 'components/Slider';
-
+import selectors from 'selectors';
+import actions from 'actions';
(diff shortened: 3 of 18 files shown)


src/components/AnnotationPopup/AnnotationPopup.js
```diff
@@ -89,6 +89,7 @@ class AnnotationPopup extends React.PureComponent {
   onAnnotationSelected = (e, annotations, action) => {
     if (action === 'selected' && annotations.length === 1) {
       const annotation = annotations[0];
+      console.log(annotation.Subject);
```
ZhijieZhang: Can you remove this console.log please?

src/components/AnnotationPopup/AnnotationPopup.js
```diff
@@ -183,6 +184,7 @@ class AnnotationPopup extends React.PureComponent {
 }
 
 const mapStateToProps = state => ({
+  test: selectors.getActiveToolName(state),
```
ZhijieZhang: This prop doesn't seem to be used in this component, can we remove this line if this is indeed not being used?

src/components/AnnotationStylePopup/AnnotationStylePopup.js
```diff
@@ -51,6 +51,7 @@ class AnnotationStylePopup extends React.Component {
     return(
       <div className={className} data-element="annotationStylePopup" onClick={() => closeElement('annotationPopup')}>
         <StylePopup 
+          annotationType = {annotation.Subject}
```
ZhijieZhang: Using `annotation.Subject` to get the annotation type is ok, but we need to be careful here :^). 
In the new UI, when an annotation is created, the value of its `Subject` depends on what language the user is using. Currently the UI supports 4 languages, which can be found under `i18n` folder, and a user can change the language by calling readerControl.setLanguage(language). 
That being said, if you called readerControl.setLanguage('fr'), and then select the highlight tool button to create an highlight annotation, the `Subject` of that annotation will be Surligne instead of Highlight. If possible, we should try to use other property than `Subject`

src/components/StylePopup/StylePopup.js
```diff
@@ -129,4 +122,13 @@ class StylePopup extends React.PureComponent {
...
+const mapStateToProps = (state,ownProps) => ({
+  colorPalette: selectors.getColorPalette(state, ownProps.annotationType),
+  annotationType: ownProps.annotationType
+});
+
+const mapDispatchToProps = {
+  setColorPalette: actions.setColorPalette,
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(StylePopup);
```
ZhijieZhang: Not sure if it would be better to connect `ColorPaletteHeader` with redux and pass `colorPalette` and `setColorPalette` to it directly instead of pass those two props through `StylePopup` (line 111)

src/components/ToolStylePopup/ToolStylePopup.js
```diff
@@ -112,6 +121,7 @@ class ToolStylePopup extends React.PureComponent {
     return (
       <div className={className} data-element="toolStylePopup" style={{ top, left }} ref={this.popup} onMouseDown={e => e.stopPropagation()} onClick={this.onClick}>
         <StylePopup
+          annotationType = {this.toolNameToAnnotationType(activeToolName)}
```
ZhijieZhang: It would be great if we can start a function name with a verb, what about `getAnnotationType` ?

src/redux/selectors/exposedSelectors.js
```diff
@@ -48,6 +48,16 @@ export const isDocumentLoaded = state => state.viewer.isDocumentLoaded;
 export const isDocumentReadOnly = state => state.viewer.isReadOnly;
 export const getCustomPanels = state => state.viewer.customPanels;
 export const getPageLabels = state => state.viewer.pageLabels;
+export const getColorPalette = (state, annotationType) => {
+  if (annotationType !== null){
+    return state.viewer.colorPalette[annotationType].currentPalette;
+  }
+};
+export const getDefaultColorPalette = (state, annotationType) => {
```
ZhijieZhang: Maybe I overlooked something, but this selector doesn't seem to be used. Is there any reason why we are having this selector? I can see `getColorPalette` is being used in `StylePopup`

JSpark22: I will be using getDefaultColorPalette to fetch the palette colour to be displayed on button icons.

ZhijieZhang:
Another PR with the latest change has been made, so this PR will be closed.

Issue #129, opened 2019-01-14
API for changing which colour to show on a button icon


ZhijieZhang:

ZhijieZhang:
This issue has been resolved in https://github.com/PDFTron/webviewer-ui/commit/712d81f9ec1d3a702462011d2e88122fc0fa1b88

PR #138, opened 2019-01-22
[update] setColorPalette API and setIconColorPalette API Implemented. 

JSpark22:
setColorPalette API changes the default color palette for an annotation.
setIconColorPalette API allows setting a default palette to import an annotation icon color from. 
Fixes issues #128 and #129.


ZhijieZhang:
Discussed with James, I will be implementing the rest of these two APIs to make them work for custom annotations/tools

PR #148, opened 2019-01-25
Implementation of new zoom dropdown UI

JSpark22:
New zoom dropdown UI is implemented.


src/components/ToggleElementDropdown/ToggleElementDropdown.js
```diff
@@ -0,0 +1,90 @@
...
+
+import ToggleElementButton from '../ToggleElementButton';
+
+import './ToggleElementDropdown.scss';
+
+class ToggleElementDropdown extends React.PureComponent {
+
+  constructor(props){
+    super(props);
+    this.state = { zoom: Math.ceil(core.getZoom() * 100), value: 100 };
```
ZhijieZhang: I'm not quite sure if we need `zoom` as part of the state. I think we can just call `setState` on `value` to trigger a re-render. What do you think?

src/components/ToggleElementDropdown/ToggleElementDropdown.js
```diff
@@ -0,0 +1,90 @@
...
+    this.textarea = React.createRef();
+  }
+  componentDidMount(){
+    core.addEventListener('zoomUpdated', this.onZoomUpdated);
```
ZhijieZhang: We are already listening this event in eventHandler.js, line 35, `core.addEventListener('zoomUpdated', onZoomUpdated);`. So every time this event is triggered, an action will be dispatched(see `onZoomUpdated.js`) to update the store. What we can do here is connect this component with redux and get the zoom value from `state.viewer.zoom`. Let me know if this is confusing.

src/components/ToggleElementDropdown/ToggleElementDropdown.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  onZoomUpdated = () => {
+    this.setState({ zoom: Math.ceil(core.getZoom() * 100)},
+    ()=>{this.setState({ value: this.state.zoom })});
+  }
```
ZhijieZhang: I think instead of updating `this.state.value` after `this.state.value` is updated, we can do something like:
```javascript
// get zoom value somehow
this.setState({ value:  Math.ceil(zoom * 100) });
```

src/components/ToggleElementDropdown/ToggleElementDropdown.js
```diff
@@ -0,0 +1,90 @@
...
+
+  onKeyPress = e => {
+    if (window.event.keyCode === 13){
+      zoomTo(this.state.value/100);
+    }
+  }
+
+  onChange = e => {
+    const re = /^[0-9\b]+$/;
+    if (re.test(e.target.value) || e.target.value === ''){
```
ZhijieZhang: It is nice to use regex here! Good job

src/components/ZoomDropdown/ZoomDropdown.js
```diff
@@ -0,0 +1,91 @@
...
+  componentDidMount() {
+    window.addEventListener('resize', this.handleWindowResize);
+  }
+
+  componentDidUpdate(prevProps) {
+    if (!prevProps.isOpen && this.props.isOpen) {
+      this.props.closeElements([ 'viewControlsOverlay', 'toolsOverlay', 'searchOverlay', 'menuOverlay', 'toolStylePopup' ]);
+      this.setState(getOverlayPositionBasedOn('zoomDropdownButton', this.dropdown),()=>{
+        this.setState({ left: this.state.left - 20 });
+      });
```
ZhijieZhang: Same here, instead of using setState twice, I think we can doing something like:
```javascript
const { left, right } = getOverlayPositionBasedOn('zoomDropdownButton', this.dropdown);
this.setState({ 
  left: left - 20,
  right
});
```
This way, this component will only re-render once.

Also I see `getOverlayPositionBasedOn` has a third parameter now, maybe we can use `getOverlayPositionBasedOn('zoomDropdownButton', this.dropdown, 'center')`?

src/components/App/App.js
```diff
@@ -26,6 +26,7 @@ import ProgressModal from 'components/ProgressModal';
 import FilePickerHandler from 'components/FilePickerHandler';
 import CopyTextHandler from 'components/CopyTextHandler';
 import PrintHandler from 'components/PrintHandler';
+import ZoomDropdown from 'components/ZoomDropdown';
```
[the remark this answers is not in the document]
JSpark22: Will do!

src/components/DropdownItem/DropdownItem.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+import './DropdownItem.scss';
+
+class DropdownItem extends React.PureComponent {
```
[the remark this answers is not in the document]
JSpark22: Done!

src/components/ToggleElementDropdown/ToggleElementDropdown.js
JSpark22: Sounds good!

src/components/ToggleElementDropdown/ToggleElementDropdown.js
JSpark22: Since zoom needs to be fetched from the core, should I modify the code to `this.setState({ value: Math.ceil(core.getZoom() * 100) });`?

src/components/ToggleElementDropdown/ToggleElementDropdown.js
JSpark22: Thank you!

src/components/ZoomDropdown/ZoomDropdown.js
JSpark22: The reason why `center` doesn't work on `getOverlayPositionBasedOn()` is because we are using custom element. Because of this, the reference becomes the textarea part of the button, instead of the whole button.

PR #208, opened 2019-03-18
[update]Expandable reply field

JSpark22:
Reply field is expanded when `NoteContents` is clicked


src/apis/enableMeasurementTool.js
```diff
@@ -0,0 +1,6 @@
+import { PRIORITY_ONE } from 'constants/actionPriority';
```
ZhijieZhang: Is this file just for testing purpose?
We already have an `enableMeasurement` API so I don't think we need this API.

src/components/NoteContents/NoteContents.js
```diff
@@ -76,7 +76,7 @@ class NoteContents extends React.Component {
     const { isEditing, closeEditing, renderContents, t, contents } = this.props;
 
     return (
-      <div className="NoteContents" onClick={e => e.stopPropagation()} onMouseDown={e => e.stopPropagation()}>
+      <div className="NoteContents" onClick={isEditing ? e => e.stopPropagation():()=>{}} onMouseDown={isEditing ? e => e.stopPropagation():()=>{}}>
```
ZhijieZhang: Yeah you are right it's getting overwritten by another commit. I noticed that in the same file we have a function `handleNoteContentsClick` which does the same thing. Do you mind using that function instead? Reason is that there's some comment which explains the intention of doing so. Thanks!

PR #226, opened 2019-03-30
[bugfix] Fix issue where calling setToolMode API with different tool names will blow up the app

ZhijieZhang:
…names will blow up the app


ZhijieZhang:
The app will blow up if we call:
```javascript
readerControl.setToolMode('AnnotationCreateArrow')
```
and then
```javascript
readerControl.setToolMode('AnnotationCreateCallout')
```

This is a bandage fix since the bug is caused by using `map.js` and it will be removed in the coming release.

PR #229, opened 2019-04-04
Add at mentions logic

elm2178:


justinjung04:
Thank you for submitting a PR!
I was informed about the at mentions API request from our solution engineers and it is in our backlog to come up with set of APIs that can be applied in different implementations.
It looks like the file permissions have changed in your commit, so I wouldn't be able to merge them to master, but we will refer to the new logics you have implemented when we work on the APIs.
I will close this PR for now, but we really appreciate your interest in WebViewer UI and effort to help everyone! 👍

PR #250, opened 2019-04-18
[BugFix] Fix bug with SignatureModal when disabling

awy3:
Added null checks to keep "getBoundingClientRect" error from happening. To reproduce error,  'documentLoaded' event call 'viewerInstance.disableElement('signatureModal')'. This error only seems to come up when developer tools aren't open.


S0ulDrag0n:
I think this issue is also resolved in #248 . That PR also fixes re-enabling an initially disabled signature modal.

PR #253, opened 2019-04-22
[bugfix] New header regression

jamesjypark:
Fixed regression regarding selectors and APIs for the new header structure


justinjung04:
- [ ] Text select tool is missing in mobile
- [ ] Default group image shouldn't change color
- [ ] Cannot delete tool buttons in root
- [ ] Should take out header items from "default"

PR #260, opened 2019-05-02
uploadData and uriData arguments

rchennafi:
Added arguments specific to our demos for uploading uris and files directly. Was formerly only available in legacy UI code.


rchennafi:
Need to redo it for dev branch.

PR #275, opened 2019-05-24
Page operation

khein-pdftron:
Fix for: https://trello.com/c/SMuRAQMV/296-page-operation-sample-is-kind-of-broken-rotate-insert-crop-and-delete-pages-will-make-other-pages-turn-blank
https://trello.com/c/uZpZWQ7A

Thumbnail.js now takes into account moved pages and added pages before deciding to reload thumbnail.


justinjung04:
Found a bug in core related to this. This PR will be merged with the core fix at the same time.

PR #277, opened 2019-05-28
Layers tab

khein-pdftron:
Only shows up when layers are available on a pdf
-added layers tab to left panel.
-can use checkbox to change visibility of each layer.


i18n/translation-en.json
```diff
@@ -68,6 +68,7 @@
     "miscToolsButton": "Misc tools",
     "notesPanel": "Notes",
     "outlinesPanel": "Outlines",
+    "layersPanel": "Layers",
```
justinjung04: Can you add the translation in other languages as well?

src/components/Layer/Layer.js
```diff
@@ -0,0 +1,41 @@
...
+
+    return (
+      <div className="Layer">
+        <Input id={layer.name} type="checkbox" label={layer.name} onChange={this.onChange} defaultChecked={layer.visible} />
+      </div>
+    );
+  }
+}
+
+export default connect(null, null)(Layer);
```
justinjung04: you can just do `export default Layer`

src/event-listeners/onDocumentLoaded.js
```diff
@@ -44,8 +44,18 @@ export default dispatch => () => {
     dispatch(actions.setOutlines(outlines));
   });
 
+  const doc = core.getDocument();
+  doc.getLayersArray().then(layers => {
+    if (layers.length === 0) {
+      dispatch(actions.disableElement('layersPanel', 1));
+      dispatch(actions.disableElement('layersPanelButton', 1));
+    } else {
+      dispatch(actions.setLayers(layers));
```
justinjung04: We should enable layersPanel and layersPanelButton here for subsequent documents

PR #286, opened 2019-06-08
Cad formats support

hbadakhshannoory:
Adding the rvt, rfa and dwf cad format to webviewer


justinjung04:
Thank you for making this PR!
Although we technically support the file types you've added, we still have work in progress and we are not ready to officially add them yet. It's likely to be part of our next release, which we don't have a solid timeline yet. We will close this PR for now, but thank you for supporting WebViewer :)

PR #298, opened 2019-06-20
Remove the use of jQuery in UI

ZhijieZhang:


src/helpers/setupLoadAnnotationsFromServer.js
```diff
@@ -20,32 +20,31 @@ export default store =>  {
       callback('');
       return;
     }
+    
+    // make sure we are not getting cached responses
+    serverUrl.indexOf('?') === -1 ? serverUrl += `?_=${Date.now()}` : serverUrl += `&_=${Date.now()}`;
+    serverUrl = documentId ? `${serverUrl}&did=${documentId}` : serverUrl;
```
justinjung04: can we turn these two lines into if conditions

PR #306, opened 2019-07-08
Fix how searching case sensitively is displayed in German ui

uluzox:
With your translation it looks like this:

d1WPymQMQzeTZ6k6

After the changes it looks like this.
FLFH56bNNq9QlYIm

I am German so, I hope you trust me with the syllable division ;-)


ZhijieZhang:
Thanks!

PR #312, opened 2019-07-19
Group annots

khein-pdftron:
Added group/ungroup buttons. Select a group of annotations and either a group or ungroup button will show in the context menu.

If there is only one group in the selection and more than one annotation, then ungroup will show.

If there is more than one group in the selection, then the group button will show. Loose annotations are considered a group of size 1.

Need to use latest changes to 5.1 branch of WebViewer for the api calls to work.


ZhijieZhang:
As discussed I've merged the new grouping UI to the 5.2 branch: https://github.com/PDFTron/webviewer-ui/tree/5.2

:)

PR #318, opened 2019-07-30
[upgrade] annotation-state-ui

aso777:
Added new functionality for annotation state. User is able to add, change, delete annotation state.

### DIFF
--- a/i18n/translation-en.json
+++ b/i18n/translation-en.json
@@ -167,6 +167,14 @@
       "precision": "Precision"
     }
   },
+  "state": {
+    "accepted": "Accepted",
+    "rejected": "Rejected",
+    "completed": "Completed",
+    "cancelled": "Cancelled",
+    "set": "Set status:",
+    "none": "None"
+  },
   "shortcut": {
     "arrow": "(A)",
     "callout": "(C)",
--- a/i18n/translation-en.json
+++ b/i18n/translation-en.json
@@ -153,6 +153,7 @@
       "completed": "Completed",
       "cancelled": "Cancelled",
       "set": "Set status:",
+      "setBy": "set by",
       "none": "None"
     },
     "measurementOverlay": {
--- a/i18n/translation-en.json
+++ b/i18n/translation-en.json
@@ -144,6 +144,17 @@
     "signatureOverlay": {
       "addSignature": "Add a signature"
     },
+    "status": {
+      "status": "Status"
+    },
+    "state": {
+      "accepted": "Accepted",
+      "rejected": "Rejected",
+      "completed": "Completed",
+      "cancelled": "Cancelled",
+      "set": "Set status:",
+      "none": "None"
+    },
     "measurementOverlay": {
       "scale": "Scale Ratio",
       "angle": "Angle",
@@ -166,14 +177,6 @@
       "page": "Page",
       "precision": "Precision"
     }
-  },
-  "state": {
-    "accepted": "Accepted",
-    "rejected": "Rejected",
-    "completed": "Completed",
-    "cancelled": "Cancelled",
-    "set": "Set status:",
-    "none": "None"
   },
   "shortcut": {
     "arrow": "(A)",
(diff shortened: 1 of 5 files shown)


i18n/translation-en.json
```diff
@@ -167,6 +167,14 @@
       "precision": "Precision"
     }
   },
+  "state": {
```
ZhijieZhang: Can we have this object in `option` please? So it will look something like this:

```
  "option": {
    "annotationColor": {
      "StrokeColor": "Border color",
      "FillColor": "Fill color",
      "TextColor": "Text color"
    },
    "displayMode": {
      "layout": "Layout",
      "pageTransition": "Page transition"
    },
    "layout": {
      "cover": "Cover",
      "double": "Double",
      "single": "Single"
    },
    ....
   "state": {
    "accepted": "Accepted",
    "rejected": "Rejected",
    "completed": "Completed",
    "cancelled": "Cancelled",
    "set": "Set status:",
    "none": "None"
  }
}
```

src/components/NotePopup/NotePopup.js
```diff
@@ -89,6 +120,7 @@ const mapStateToProps = state => ({
   isDisabled: selectors.isElementDisabled(state, 'notePopup'),
   isEditDisabled: selectors.isElementDisabled(state, 'notePopupEdit'),
   isDeleteDisabled: selectors.isElementDisabled(state, 'notePopupDelete'),
+  isStateDisabled: selectors.isElementDisabled(state, 'notePopopState'),
```
ZhijieZhang: I think it should be `notePopupState` instead?

src/components/NoteRoot/NoteRoot.js
```diff
@@ -68,10 +69,14 @@ class NoteRoot extends React.Component {
 
   render() {
     const { annotation, renderContents, isEditing, closeEditing, searchInput, contents } = this.props;
+    const annotationState = annotation.getStatus();
 
     return (
       <div className="NoteRoot">
         {this.renderHeader()}
+        {annotationState && annotationState !== '' && annotationState !== 'None' &&
```
ZhijieZhang: I think if `annotationState` is truthy then it implies it won't equal to `''` so we can just have 

```javascript
annotationState && annotationState !== 'None' && ...
```

src/components/NoteRoot/NoteRoot.js
```diff
@@ -68,10 +69,14 @@ class NoteRoot extends React.Component {
...
+          <div className="status">Status: {annotationState}</div>
```
ZhijieZhang: Do we also want to translate the `Status` word?

src/components/NotePopup/NotePopup.js
```diff
@@ -57,16 +58,37 @@ class NotePopup extends React.Component {
     this.props.openEditing();
   }
 
+  onAccept = () => {
+    core.getAnnotationManager().updateAnnotationState(this.props.annotation, 'Accepted');
```
ZhijieZhang: Can we combine these functions into one? Something like this:

```javascript
onStateUpdate = state => {
  core.getAnnotationManager().updateAnnotationState(this.props.annotation, state);  
}
```

and in the jsx below you can do something like:
```javascript
<div data-element="notePopupStateAccepted" onClick={() => this.onStateUpdate('Accept')}>{t('state.accepted')}</div>
```

The drawback of this is that this div will re-render every time when this component re-renders since we are creating a new function. But I think this is okay. Let me know if you have any concerns

i18n/translation-en.json
aso777: Resolved in dcb321c. Thank you!

src/components/NotePopup/NotePopup.js
aso777: Resolved in dcb321c. Thank you!

src/components/NoteRoot/NoteRoot.js
aso777: Resolved in dcb321c. Thank you!

src/components/NoteRoot/NoteRoot.js
aso777: Resolved in dcb321c. Thank you!

src/components/NotePopup/NotePopup.js
aso777: Resolved in dcb321c. Thank you!

PR #322, opened 2019-08-07
[update] change to getFileData

sisardor:
change as per request

### DIFF
--- a/src/helpers/downloadPdf.js
+++ b/src/helpers/downloadPdf.js
@@ -28,20 +28,18 @@ export default (dispatch, options) => {
     const downloadName = getDownloadFilename(name, '.pdf');
 
     const doc = core.getDocument();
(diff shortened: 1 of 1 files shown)


src/helpers/downloadPdf.js
```diff
@@ -28,7 +28,7 @@ export default (dispatch, options) => {
     const downloadName = getDownloadFilename(name, '.pdf');
 
     const doc = core.getDocument();
-    const bbURLPromise = externalURL ? Promise.resolve({ url: externalURL }) : doc.getDownloadLink({ filename: downloadName });
+    const bbURLPromise = externalURL ? Promise.resolve({ url: externalURL }) : doc.getFileData({ filename: downloadName });
```
mparizeau-pdftron: I think for this change we can't just change getDownloadLink with getFileData because one returns a URL and the other returns an ArrayBuffer. I think we may need to keep the externalURL for redacted file downloading and the else case it uses getFileData.

So down below instead of `if (bbURLPromise)` maybe we have `if (externalURL)`.

PR #323, opened 2019-08-09
[Bugfix] Fix for allowing the left panel to be adjustable in IE

awy3:
Since we were using CSS variables (not supported in IE) the left panel wasn't adjustable in IE. I updated  it to use JavaScript for styling in IE ( it still use CSS for other broswers)

### DIFF
--- a/src/components/LeftPanel/LeftPanel.js
+++ b/src/components/LeftPanel/LeftPanel.js
@@ -22,8 +22,16 @@ class LeftPanel extends React.Component {
     isOpen: PropTypes.bool,
     customPanels: PropTypes.array.isRequired,
     activePanel: PropTypes.string.isRequired,
-    closeElement: PropTypes.func.isRequired
+    closeElement: PropTypes.func.isRequired,
   };
+
+  constructor() {
+    super();
+    this.state = {
+      // IE11 doesn't support css variables, so use the hard coded value from App.scss
+      width: !isIE11 ? document.body.style.getPropertyValue('--left-panel-width') : 300
+    };
+  }
 
   componentDidUpdate(prevProps) {
     if (!prevProps.isOpen && this.props.isOpen && isTabletOrMobile()) {
@@ -35,13 +43,29 @@ class LeftPanel extends React.Component {
     return panel === this.props.activePanel ? 'flex' : 'none';
   };
 
+  onWidthChange = width => {
+    this.setState({ width });
+  }
+
   render() {
     const { isDisabled, closeElement, customPanels } = this.props;
-    
+    const { width } = this.state;    
     if (isDisabled) {
       return null;
     }
-    
+
+    let style = { };
+    let innerStyle = { };
+
+    if (isIE11) {
+      // IE11 will use javascript for controlling width, other broswer will use CSS
+      style = width ? { width } : null;
+
+      innerStyle = {
+        left: width
+      };
+    }
+
     const className = getClassName('Panel LeftPanel', this.props);
 
     return (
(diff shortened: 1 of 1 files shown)


src/components/LeftPanel/LeftPanel.js
```diff
@@ -35,13 +38,30 @@ class LeftPanel extends React.Component {
     return panel === this.props.activePanel ? 'flex' : 'none';
   };
 
+  onWidthChange = width => {
+    if (isIE11 || isIEEdge) {
+      this.props.updateLeftPanel(width);
```
ZhijieZhang: Do you think naming the function as `updateLeftPanelWidth` or `setLeftPanelWidth` would be better options?

ZhijieZhang: Sorry.. And do you think we need some comments here? Explain why in IE and Edge we are updating the width in redux while in other browsers we don't

src/components/LeftPanel/LeftPanel.js
```diff
@@ -35,13 +38,30 @@ class LeftPanel extends React.Component {
...
+
+    let style = { };
+    let innerStyle = { };
+
+    if (isIE11) {
+      // IE11 will use javascript for controlling width, other broswers will use CSS
+      style = leftPanelWidth ? { width: leftPanelWidth } : null;
+
+      innerStyle = {
+        left: leftPanelWidth
```
ZhijieZhang: I don't remember the reason why we are using `left` for absolutely positioning the resize bar.

In `LeftPanel.scss`, we could use `right: 0px` and `margin-right: -10px` and probably get rid of this.

src/components/LeftPanel/LeftPanel.js
```diff
@@ -61,7 +82,7 @@ class LeftPanel extends React.Component {
           <LeftPanelTabs />
         </div>
 
-        <ResizeBar />
+        <ResizeBar onWidthChange={this.onWidthChange} style={innerStyle}/>
```
ZhijieZhang: Use `right: 0` for the ResizeBar so we don't need to pass a style prop down to it

src/helpers/documentContainerHelper.js
```diff
@@ -25,6 +25,9 @@ export const updateContainerWidth = (prevProps, props, container) =>  {
   if (leftPanelOpened) {
     shrinkContainerWidthBy(leftPanelWidth, container);
     container.style.marginLeft = `${leftPanelWidth}px`;
+  } else if (props.isLeftPanelOpen) {
+    container.style.marginLeft = `${leftPanelWidth}px`;
+    container.style.width = `calc(100% - ${leftPanelWidth}px)`;
```
ZhijieZhang: Is moving the resize bar working fine on your IE or Edge?
I tried on my computer and when I move the resize bar, I saw the position for the scroll bar is shifted as well. I suspect it's because the width of the documentContainer isn't correct?

src/components/LeftPanel/LeftPanel.js
awy3: Your right it should be 'setLeftPanelWidth'. Also I will add a comment explaining why we are doing this for IE and Edge

src/components/LeftPanel/LeftPanel.js
awy3: Using CSS for this is better. Before it was using CSS variable for "left" but since that wasn't supported in IE11, I made it set "left" with JavaScript instead. I tested with `right: 0px` and it seem to work in IE

src/components/LeftPanel/LeftPanel.js
awy3: I updated it to use `right: 0`

src/helpers/documentContainerHelper.js
awy3: It looks like the css transition is causing the scroll bar to jump around. I disable it for IE and Edge and it seem to fix the issue (make resize not as smooth in IE and Edge but the scroll bar jumping is gone ).

PR #324, opened 2019-08-09
[bugfix] Deleting pages with annotations on the last page fix

awy3:
There is a bug where WebViewer crashes when deleting pages with an annotation on the last page. You can see it at https://www.pdftron.com/samples/web/samples/pdf-manipulation/page-operations/  (create some annotations on the last page and delete the first page).

The problem is from a race condition when calling 'setPageLabels'. The note panel uses 'sortStrategies.js' which calls "getDocumentCenter(pageNumber)". However the 'pageNumber' isn't updated yet, so the annotation on the last page will cause an exception. I added 'setTimeout' so that it'll wait for the pageNumbers to update first.


src/event-listeners/onLayoutChanged.js
```diff
@@ -8,8 +8,8 @@ export default dispatch => (e, { added, removed }) => {
     dispatch(actions.setTotalPages(totalPages));
 
     setTimeout(() => {
-      // this 'onLayoutChange' handler get trigger before the 'onLayoutChange' handler on 'viewerPageManager' get fire (so page number aren't updated yet)
-      // use 'setTimeout' so the annotationManager.onLayoutChange handler finish before we call 'setPageLabels'
+      // this 'onLayoutChange' handler get trigger before the 'onLayoutChange' handler on 'viewerPageManager' get trigger (so page numbers aren't updated yet)
+      // use 'setTimeout' so the other 'onLayoutChange' handler finishs before we call 'setPageLabels'
```
ZhijieZhang: I think it should read: this 'onLayoutChange' handler get(s) trigger(ed).

It should be annotManager instead of 'viewerPageManager' I think.

finishs -> finishes?

awy3: I updated the comment

PR #326, opened 2019-08-14
[Feature] Added API for setting max number of signatures

awy3:
Added an API for signature count. I'm not too sure about the "onSaveDefault" function in the SignatureOverlay.js file


src/apis/setMaxSignaturesCount.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Set the number of signatures that can be store in the WebViewer (default is 2)
```
ZhijieZhang: Should this be `stored`?

awy3: You're correct. I need to add a grammar check plugin for VS code when I have time

PR #334, opened 2019-08-20
[update] Annotation Overlay UI Component 

jamesjypark:
Annotation overlay component added to display text content of annotations when hovering. 
The overlay will display the author, the text content and the number of replies when hovering an annotation with text contents. 

Annotation Overlay can be disabled using its `dataElement`:
```
readerControl.disableElement('annotationOverlay')
```
Likewise, it can be re-enabled by the following command:
```
readerControl.enableElement('annotationOverlay')
```


src/components/AnnotationOverlay/AnnotationOverlay.js
```diff
@@ -0,0 +1,76 @@
...
+    left: 0,
+    top: 0
+  });
+  
+  const am = core.getAnnotationManager();
+  
+  useEffect(() => {
+    window.docViewer.on('mouseMove.hover', onMouseHover);
+    return () => {
+      window.docViewer.off('mouseMove.hover', onMouseHover);
```
ZhijieZhang: I don't think we need to use event namespace here. Doing `mouseMove` should be OK. Can we also use `core.addEventListener` just for consistency?

src/components/AnnotationOverlay/AnnotationOverlay.js
```diff
@@ -0,0 +1,76 @@
...
+  const onMouseHover = (e, mouseEvent) => {
+    const annotation = am.getAnnotationByMouseEvent(mouseEvent);
+    setIsOpen(!!annotation);
+    if (annotation) {
+      setAnnotation(annotation);
+      setOverlayPosition({
+        left: mouseEvent.clientX + 20,
+        top: mouseEvent.clientY + 20
+      });
+      setRepliesCount(annotation.getReplies().length);
```
ZhijieZhang: Can you move this function into the `effect` hook? If the this handler is using any props or state, being in the same function would be clearer to see what dependencies we need to pass to the `useEffect`

src/components/AnnotationOverlay/AnnotationOverlay.js
```diff
@@ -0,0 +1,76 @@
...
+const MAX_WORDS = 100;
+
+const AnnotationOverlay = () => {
+  const isDisabled = useSelector(state => 
+    selectors.isElementDisabled(state, 'annotationOverlay')
+  );
+  const [isOpen, setIsOpen] = useState(false);
+  const [annotation, setAnnotation] = useState();
+  const [repliesCount, setRepliesCount] = useState(0);
+  const [overlayPosition, setOverlayPosition] = useState({
```
ZhijieZhang: Do we need `isOpen` and `repliesCount` states? If annotation is null we can just not render this component. And in the jsx we can use `annotation.getReplies()` to get the replies

src/components/AnnotationOverlay/AnnotationOverlay.js
jamesjypark: Sure, sounds good!

src/components/AnnotationOverlay/AnnotationOverlay.js
```diff
@@ -0,0 +1,72 @@
...
+        style={{ left, top }}
+      >
+        <div className="author">
+          {annotation.Author}
+        </div>
+        <div>
+          {contents}
+        </div>
+        <div>
+          {repliesCount > 0 && `Replies (${repliesCount})`}
```
ZhijieZhang: Sorry James, I overlooked this. I think we'd want to internationalize `Replies`. In `translation-en.json` it has a `action.reply`.

src/components/AnnotationOverlay/AnnotationOverlay.js
```diff
@@ -0,0 +1,72 @@
...
+      core.removeEventListener('mouseMove', onMouseHover)
+    }
+  }, []);
+
+  if (!isDisabled && annotation && annotation.getContents()) {
+    const { left, top } = overlayPosition;
+    const repliesCount = annotation.getReplies().length || 0;
+    let contents = annotation.getContents();
+    // display upto MAX_CHARACTERS characters
+    if (contents && contents.length > MAX_CHARACTERS) {
```
ZhijieZhang: `contents` should be defined here so I think we can remove this check.

ZhijieZhang:
@jamesjypark Hey, I'm gonna make some small changes to the component.

It's actually not necessarily to use `connect` for it because we are using hooks from `react-redux`. And `react-i18next` has a hook version so instead of using `withTranslation` I'm going to use `useTranslation` inside the component.

Also I think we are going to make this component off by default, and users can show it by `enableElements`.

After these changes I will merge this PR to dev, and pick to 5.2. Thanks for your contribution James :)

PR #343, opened 2019-08-28
[Bugfix] Fix problem with "goToNextPage" and "goToPrevPage" when showing more then one page

awy3:
Calling "goToPrevPage" or "goToNextPage" changes the current page by 1. This is a problem when showing two pages at once because it doesn't update the display pages.

Also, I noticed that sometimes it'll do the correct thing and change the page by 2 while other times it doesn't. This seems to be a bug

This might be a core change, current when you zoom in, it'll focus on the "next" or "prev" page. We might want to make it increment/decrement by 2 when it's zoom out while keeping the current behaviour it has now.

### DIFF
--- a/src/apis/goToNextPage.js
+++ b/src/apis/goToNextPage.js
@@ -32,11 +32,20 @@ import selectors from 'selectors';
 export default store => () => {
   const state = store.getState();
   const currentPage = selectors.getCurrentPage(state);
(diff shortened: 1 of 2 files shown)


ZhijieZhang:
Discussed with Andrew and we'll close this PR for now.

PR #344, opened 2019-08-28
Upgrade/Measurement Tool Improvements

aso777:
- Switched to controlled inputs,
- Formatting of digits (, , .) based on language
- Styling changes
- Code refactor

Unit tests
- ensured the mobile mode works
- ensured the tool properties and style updates work


aso777:
Before
Screen Shot 2019-08-28 at 3 56 17 PM

After
Screen Shot 2019-08-28 at 3 54 40 PM

aso777:
Screen Shot 2019-08-28 at 3 58 53 PM

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -2,14 +2,23 @@ import React from 'react';
...
+    super(props);
+    this.state = { currScaleFrom: props.scale[0][0],
+      currUnitFrom: props.scale[0][1],
+      currScaleTo: props.scale[1][0],
+      currUnitTo: props.scale[1][1],
+      currPrecision: props.precision,
+      isEditing: false };
+  }
+
   static propTypes = {
```
ZhijieZhang: Could we move propTypes to the top of the class just like other components?

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -2,14 +2,23 @@ import React from 'react';
...
 
 import selectors from 'selectors';
 
 import './MeasurementOption.scss';
 
-class MeasurementOption extends React.PureComponent {
+class MeasurementOption extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { currScaleFrom: props.scale[0][0],
```
ZhijieZhang: Can we have properties in separate lines like:
```javascript
this.state = {
  currScaleFrom: props.scale[0][0],
  currUnitFrom: props.scale[0][1],
  currScaleTo: props.scale[1][0],
  currUnitTo: props.scale[1][1],
  currPrecision: props.precision,
  isEditing: false
};
```

Not you fault... We should run linting in the CI for the UI as well

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -35,59 +44,71 @@ class MeasurementOption extends React.PureComponent {
...
+      this.props.onStyleChange('Precision', this.state.currPrecision);
+    });
+  }
 
-    scaleFrom = scaleFrom || scale[0][0];
-    unitFrom = unitFrom || scale[0][1];
-    scaleTo = scaleTo || scale[1][0];
-    unitTo = unitTo || scale[1][1];
+  onUnitChange = (event, type) => {
+    this.setState({ [type]: event.target.value }, () => {
```
ZhijieZhang: I feel like we could combine `onUnitChange` together with `onScaleChange` as the setState callback is exactly the same. But I'm good with the code now, easy to understand.

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -35,59 +44,71 @@ class MeasurementOption extends React.PureComponent {
...
-    const scaleToRefValue = this.scaleToRef.current.value;
-    const [[scaleFrom], [scaleTo]] = this.props.scale;
+    return lang;
+  }
 
-    if (scaleFromRefValue === '') {
-      this.scaleFromRef.current.value = scaleFrom;
-    } else if (scaleToRefValue === '') {
-      this.scaleToRef.current.value = scaleTo;
+  formatValue = value => {
```
ZhijieZhang: Can we use the same coding style as `getLanguage` for this function? Something like:
```javascript
  formatValue = value => {
    const lang = this.getLanguage();

    if (lang === 'de') {
      value = value.toLocaleString('de-DE', { maximumFractionDigits: 4 });
    } else if (lang === 'fr') {
      value = value.toLocaleString('fr-FR', { maximumFractionDigits: 4 });
    } else if (lang === 'ru') {
      value = value.toLocaleString('ru-RU', { maximumFractionDigits: 4 });
    }

    return value;
  }
```

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -35,59 +44,71 @@ class MeasurementOption extends React.PureComponent {
...
+    } else if (lang === 'ru') {
+      return value.toLocaleString('ru-RU', { maximumFractionDigits: 4 });
     }
-  };
+    
+    return value;
+  }
+
+  toggleEditing = () => {
+    this.setState({ isEditing: !this.state.isEditing });
```
ZhijieZhang: Can we use the callback updater for this `setState` please. 
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -35,59 +44,71 @@ class MeasurementOption extends React.PureComponent {
...
+  }
 
   render() {
     const {
-      scale,
-      precision,
       measurementUnits,
-      openMeasurementDropdown,
       onOpenDropdownChange,
```
ZhijieZhang: I don't think we need this prop anymore

src/components/MeasurementOption/MeasurementOption.js
aso777: One is doing is converting a value to a number and the other one is doing string. Would have to write if check, but for now its easy to read and it is separated.

src/components/MeasurementOption/MeasurementOption.js
aso777: Didn't seem like we needed it. Appears to be leftovers from the previous implementation.

src/components/MeasurementOption/MeasurementOption.js
aso777: Resolved in 084caf3. Thanks!

src/components/MeasurementOption/MeasurementOption.js
aso777: Resolved in 084caf3. Thanks!

src/components/MeasurementOption/MeasurementOption.js
aso777: Resolved in 084caf3. Thanks!

src/components/MeasurementOption/MeasurementOption.js
aso777: Resolved in 084caf3. Thanks!

src/components/MeasurementOption/MeasurementOption.js
aso777: Resolved in 084caf3. Thanks!

src/components/MeasurementOption/MeasurementOption.js
aso777: Resolved in 084caf3. Thanks!

PR #346, opened 2019-08-30
[Bugfix]Added missing "LockContents" support

awy3:
We were missing support for "LockedContents".  The following is what it looked like in the legacy UI
image

### DIFF
--- a/src/components/NotePopup/NotePopup.js
+++ b/src/components/NotePopup/NotePopup.js
@@ -32,6 +32,7 @@ const NotePopup = ({ annotation, setIsEditing }) => {
     shallowEqual,
   );
   const [canModify, setCanModify] = useState(core.canModify(annotation));
+  const [canModifyContents, setCanModifyContents] = useState(core.canModifyContents(annotation));
   const [t] = useTranslation();
   const dispatch = useDispatch();
   const isOpen = notePopupId === annotation.Id;
@@ -39,6 +40,7 @@ const NotePopup = ({ annotation, setIsEditing }) => {
   useEffect(() => {
     const onUpdateAnnotationPermission = () => {
       setCanModify(core.canModify(annotation));
+      setCanModifyContents(core.canModifyContents(annotation));
     };
 
     core.addEventListener(
@@ -72,7 +74,10 @@ const NotePopup = ({ annotation, setIsEditing }) => {
     core.deleteAnnotations([annotation]);
   };
 
-  return !canModify || isDisabled ? null : (
+  const isEditable = !isEditDisabled && canModifyContents;
+  const isDeletable = !isDeleteDisabled && canModify;
+
+  return !(isEditable || isDeletable) || isDisabled ? null : (
     <div
       className="NotePopup"
       data-element="notePopup"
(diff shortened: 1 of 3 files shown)


src/components/NotePopup/NotePopup.js
```diff
@@ -83,12 +88,12 @@ const NotePopup = ({ annotation, setIsEditing }) => {
       </div>
       {isOpen && (
         <div className="options" onClick={closePopup}>
-          {!isEditDisabled && (
+          {isEditable && (
```
ZhijieZhang: Sorry, not sure if I missed something, but can't we have `!isEditDisabled && canModifyContents && (...)`?
Why do we need `isEditable` and `isDeletable`?

```javascript
return !canModify || isDisabled ? null : (
  <div
    className="NotePopup"
    data-element="notePopup"
    onMouseDown={e => e.stopPropagation()}
  >
    <div className="overflow" onClick={togglePopup}>
      <Icon glyph="ic_overflow_black_24px" />
    </div>
    {isOpen && (
      <div className="options" onClick={closePopup}>
        {!isEditDisabled && canModifyContents && (
          <div data-element="notePopupEdit" onClick={handleEdit}>
            {t('action.edit')}
          </div>
        )}
        {!isDeleteDisabled && (
          <div data-element="notePopupDelete" onClick={handleDelete}>
            {t('action.delete')}
          </div>
        )}
      </div>
    )}
  </div>
```

awy3: You mean do something like 
<pre>
 return !(  <b>!(isEditDisabled && canModifyContents) || !(isDeleteDisabled && canModify)</b> ) || isDisabled ? null : (
    < div
      className="NotePopup"
      data-element="notePopup"
      onMouseDown={e => e.stopPropagation()}
    >
      <  div className="overflow" onClick={togglePopup}>
        <Icon glyph="ic_overflow_black_24px" />
      </  div>
      {isOpen && (
        <  div className="options" onClick={closePopup}>
          {  <b> isEditDisabled && canModifyContents </b> && (
            <  div data-element="notePopupEdit" onClick={handleEdit}>
              {t('action.edit')}
            </  div>
          )}
          {  <b> isDeleteDisabled && canModify </b> && (
            <  div data-element="notePopupDelete" onClick={handleDelete}>
              {t('action.delete')}
            </  div>
          )}
        </  div>
      )}
    </div>
  );
</pre>

ZhijieZhang: Nah I just meant the code in my response.
Is just changing 
```javascript
        {!isEditDisabled && (
          <div data-element="notePopupEdit" onClick={handleEdit}>
            {t('action.edit')}
          </div>
        )}
```
to
```javascript
        {!isEditDisabled && canModifyContents && (
          <div data-element="notePopupEdit" onClick={handleEdit}>
            {t('action.edit')}
          </div>
        )}
```
wrong?

ZhijieZhang: Why do we want to change `!canModify || isDisabled ?` to something else?

awy3: "isEditable" is made up of  "!isEditDisabled && canModifyContents" already.  I guess the name is similar to "isEditDisabled" (which is when the dataElement is disabled)

awy3: I'm using isEditable and isDeletable as temp variables because we are also replacing 
```javaScript
 return !canModify || isDisabled ? null : (
```
with 
```javaScript
 return !(isEditable || isDeletable) || isDisabled ? null : (
```

ZhijieZhang: yeah this is the part I don't understand. I'm ok with having the `isEditable` variable, but I don't understand why we are replacing `!canModify` with `!(isEditable || isDeletable)`. Apparently they are not equivalent and I'm not sure if I'm missing something here.

awy3: Actually let have a discussion about this later. I think the legacy UI comment section (above) might be doing something incorrectly, I want to confirm what the proper behavior it is for us to handle this.

PR #347, opened 2019-08-30
Feature/watermark printing

achenPDFTron:


i18n/translation-en.json
```diff
@@ -213,5 +215,12 @@
     "pan": "Pan",
     "select": "Select",
     "Marquee": "Marquee Zoom"
+  },
+  "watermark": {
```
ZhijieZhang: Can you move this inside the `option` object please.

src/components/PrintModal/PrintModal.js
```diff
@@ -41,20 +42,29 @@ class PrintModal extends React.PureComponent {
     this.includeComments = React.createRef();
     this.pendingCanvases = [];
     this.state = {
+      allowWatermarkModal: false,
```
ZhijieZhang: What's this state used for? If an user has programmatically called `setWatermark` then we don't want to show the edit watermark option?

achenPDFTron: correct.

src/redux/selectors/exposedSelectors.js
```diff
@@ -99,7 +99,9 @@ export const isEmbedPrintSupported = state => {
     window.navigator.userAgent.indexOf('Chrome') > -1 &&
     window.navigator.userAgent.indexOf('Edge') === -1;
   const isPDF = getDocumentType(state) === workerTypes.PDF;
-  return isPDF && isChrome && state.viewer.useEmbeddedPrint;
+  // return isPDF && isChrome && state.viewer.useEmbeddedPrint;
+  // TODO remove this later
+  return false;
```
ZhijieZhang: Need to revert this change.

src/helpers/print.js
```diff
@@ -3,7 +3,7 @@ import i18n from 'i18next';
 import actions from 'actions';
 import core from 'core';
 
-export default (dispatch, isEmbedPrintSupported) => {
+export default (dispatch, isEmbedPrintSupported) =>  {
```
ZhijieZhang: Hmmm what happened here?

achenPDFTron: was it supposed to be removed? i think i must have had some merge conflict problems and I might have resolved it incorrectly ?

src/event-listeners/onDocumentLoaded.js
```diff
@@ -12,12 +12,10 @@ export default dispatch => () => {
   dispatch(actions.setDocumentLoadingProgress(1));
   dispatch(actions.setWorkerLoadingProgress(1));
 
-  setTimeout(() => {
-    dispatch(actions.closeElement('progressModal'));
-    dispatch(actions.resetLoadingProgress());
-    dispatch(actions.resetUploadProgress());
-    dispatch(actions.setIsUploading(false));
-  }, 300);
+  dispatch(actions.closeElement('progressModal'));
```
ZhijieZhang: I feel like we need the 300ms timeout. If we remove this I see that the progress bar is not filled before closing...

test/cypress/integration/watermarkModal.spec.js
```diff
@@ -0,0 +1,331 @@
+const ID = {
```
ZhijieZhang: Let's skip these tests until 6.0 is out so that we don't need to commit the `lib` folder to git.

Can you remove the lib folder and add it back to the .gitignore please?

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -0,0 +1,434 @@
...
+    modalClosed: PropTypes.func,
+    formSubmitted: PropTypes.func,
+    t: PropTypes.func.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    const locationSettings = this.initializeLocationSettings();
+    this.state = {
+      isVisible: false,
```
ZhijieZhang: I don't really think we need this state because `this.state.isVisible` isn't used in the render function or anywhere else. I think we can just use `this.props.isVisible` so that this component is fully controlled by the `PrintModal`? What do you think?

i18n/translation-en.json
achenPDFTron: i have moved it in

src/redux/selectors/exposedSelectors.js
achenPDFTron: i will revert it as the very last thing before i merge.
I want to be able to test out any new code changes suggested in this PR to make sure it doesn't break.

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -0,0 +1,434 @@
...
+      isColorPaletteVisible: false,
+      // eslint-disable-next-line object-shorthand
+      locationSettings: locationSettings,
+      previousLocationSettings: locationSettings,
```
ZhijieZhang: I'm not very clear about why we need this state. Is this related to the `watermarkModalOption` state in the PrintModal?

If yes can we use `watermarkModalOption` instead?

test/cypress/integration/watermarkModal.spec.js
achenPDFTron: right now. tests can be ran on pipeline but they can't be ran locally, and i'm having trouble setting up docker locally. sure. i am fine with this.

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: good catch.
i agree with you about removing "isVIsible" from the state because its not being used.
Since it is not being used at all, it might be confusing to other devs if we kept it there.

achenPDFTron: i have removed it.

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: when we press "Ok". those form field settings need to be saved.
so that next time we open up the watermark modal, we can retrieve those saved settings.

We need it for this other case:
1. user fools around with the form field settings but does not press ok and close the dialog.
2. user re-opens watermark modal. it should be able to display the previous saved setting.

let me know if that is unclear.

src/helpers/print.js
achenPDFTron: as discussed. it might have been a bad resolving of a merge conflict. i have reverted it back to the original.

src/event-listeners/onDocumentLoaded.js
achenPDFTron: put back in the 300 ms timeout. ran end to end tests again. they aren't affected.

test/cypress/integration/watermarkModal.spec.js
achenPDFTron: should i comment out the end to end test phase in the pipeline?

src/redux/selectors/exposedSelectors.js
achenPDFTron: @mparizeau-pdftron when you think everything is good. i will fix this line.

src/components/PrintModal/PrintModal.js
```diff
@@ -446,40 +485,21 @@ class PrintModal extends React.PureComponent {
...
-            {isPrinting ? (
-              <div className="button" onClick={this.cancelPrint}>
-                {t('action.cancel')}
-              </div>
-            ) : (
-              <div className="button" onClick={this.closePrintModal}>
-                {t('action.close')}
-              </div>
-            )}
+            <button className="button" onClick={event => this.createPagesAndPrint(event)} disabled={count > -1}>{t('action.print')}</button>
```
mparizeau-pdftron: Why do we need to make this an inline function, can we do `onClick={this.createPagesAndPrint}` instead?

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -0,0 +1,429 @@
...
+        if (nodes && nodes.length > 0) {
+          this.canvasContainerRef.current.removeChild(nodes[0]);
+        }
+        canvas.style.border = '1px solid black';
+        canvas.style.height = this.canvasContainerRef.current.style.height;
+        this.canvasContainerRef.current.appendChild(canvas);
+      },
+    });
+
+    // Note: do not update and refresh the doc else it may affect other docs as well
```
mparizeau-pdftron: What does this comment mean?

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -0,0 +1,429 @@
...
+
+  createWatermarks() {
+    let watermarks = {};
+
+    Object.keys(WATERMARK_LOCATIONS).forEach(key => {
+      const temp = this.constructWatermarkOption(this.state.locationSettings[key]);
+
+      switch (WATERMARK_LOCATIONS[key]) {
+        case WATERMARK_LOCATIONS.CENTER:
+          watermarks = {
```
mparizeau-pdftron: Is there a reason this needs to be a new object reference on every iteration of the forEach? Can we just do `watermarks.diagonal = temp;` instead? Or even create a map between WATERMARK_LOCATIONS and the names of the properties on the API?

src/components/PrintModal/PrintModal.js
achenPDFTron: I don't think I may be on the same page as you on the definition of "inline function".

Here is what I see from Googling: 
image
https://stackoverflow.com/questions/19159703/what-is-the-difference-between-anonymous-and-inline-functions-in-javascript

I believe you are asking why I'm choosing to use an arrow function over not using an arrow function and doing this instead: {this.createPagesAndPrint} ?

By doing {this.createPagesAndPrint}, I'm seeing that "this" is undefined inside that function when it is called.
image

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: Maybe its a poor comment on my end.

What i'm trying to say is that, developers should not add the following 2 lines:
window.docViewer.refreshAll();
window.docViewer.updateView();

Or else, the watermark can be seen in the other document:
image


I guess I should reword the comment for clarity?

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: I don't have a strong reason why I keep creating a new object reference on each iteration of the for each. Its probably something I thought of at the spur of that moment when implementing.

I like your suggestion of creating a map between WATERMARK_LOCATIONS and the names of the properties on the API. It cuts down on the huge switch case block that you see. I have changed the code to reflect that.

src/components/PrintModal/PrintModal.js
achenPDFTron: as discussed. i have made majority of the code i 've written to match what is in the code base.

src/redux/selectors/exposedSelectors.js
achenPDFTron: @mparizeau-pdftron i have reverted this line back to what it used to be.

PR #360, opened 2019-09-25
Update MeasurementOption.js

aso777:
- Fixed an issue with Firefox firing additional onBlur after onFocus preventing user entry of scale.

### DIFF
--- a/src/components/MeasurementOption/MeasurementOption.js
+++ b/src/components/MeasurementOption/MeasurementOption.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
+import { isFirefox } from 'helpers/device';
 import i18next from 'i18next';
 
 import selectors from 'selectors';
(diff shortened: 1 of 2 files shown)


src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -114,6 +114,51 @@ class MeasurementOption extends React.Component {
     this.setState(state => ({ isEditing: !state.isEditing }));
   };
 
+  renderScaleInput = (type, val) => {
+    /**
+     * There is a bug with Firefox 69 where after onFocus, it calls onBlur right away. Remove after the issue resolved.
+     */
+    const isFirefox = typeof InstallTrigger !== 'undefined';
```
ZhijieZhang: I can verify the issue is fixed!

I have one question about this line. I'm not sure why we are using `InstallTrigger` to check if it's FF.
Can we use `navigator.userAgent` to check the type of browser? Also, could you move this to `device.js`, just like `isIE11`. Thanks!

aso777: Thank you! Resolved in a2be101.

PR #370, opened 2019-10-08
Bug fix/tooltip text overflow

achenPDFTron:


achenPDFTron:
Solution came from https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container.

It looked like we were missing one line:
word-break: break-all;
word-break: break-word; // not supported in Edge

achenPDFTron:
Picture of what the fix looks like (working on IE11, FF, Edge, Chrome).

Original problem:
https://trello.com/c/iP1tEZ6v/532-tooltip-text-overflows-improperly-for-wrong-text

Using only the letter "a"
image


Using 10 paragraphs of lorem ipsum:
image

ZhijieZhang:
@achenPDFTron  There's one lint error. Could you please fix it? I don't think you introduced it though...

achenPDFTron:
@ZhijieZhang addressed linter problem

PR #381, opened 2019-10-16
[Bugfx] Fixed issue with document drifting around in IE and Edge

awy3:
If a user clicks the top right note panel () or programmatically use "readerControl.setSideWindowVisibility()" quickly, the document with drift around.  You don't see this if you use it like a normal person

20191016_DocumentDrifting

You can reproduce it here 
https://www.pdftron.com/samples/web/samples/viewing/viewing/
What it does depends on your zoom level and the size of the browser window. 

UPDATE: read below comment


awy3:
need to test something first

awy3:
I thought this fix worked, it was fine in "localhost:3000" but after I ran "npm run build" and tested it, I notice issues

awy3:
I found the problem, the scrollbarwidth = 17, wasn't affecthing anything when I was using localhost:3000 but did after `npm run build`

PR #386, opened 2019-10-24
[Feature] hotkey: space

jiaqihuang-233:


src/helpers/hotkeysManager.js
```diff
@@ -246,6 +250,19 @@ WebViewer(...)
...
         }
       },
+      /**
+       * Hold to switch between AnnotationEdit mode and Pan Mode
+       * @name WebViewer.Hotkeys#Space
+       */
+      space: e => {
+        e.preventDefault();
+        if (e.type === 'keyup') {
+          setToolModeAndGroup(store, 'AnnotationEdit', '');
```
ZhijieZhang: Hmm I think what we want is to switch to Pan temporarily. For example, if you are using a freehand tool, you can switch to the Pan tool by holding the space, but after you release the space key the tool should be switched back to the freehand tool. 

You can check the behaviour in the legacy UI here https://www.pdftron.com/samples/web/samples/viewing/viewing-with-legacy-ui/

src/helpers/hotkeysManager.js
```diff
@@ -65,7 +65,11 @@ WebViewer(...)
     }
 
     // https://github.com/jaywcjlove/hotkeys#defining-shortcuts
-    hotkeys(key, handler);
+    if (key === 'space') {
+      hotkeys(key, { keyup: true }, handler);
+    } else {
+      hotkeys(key, handler);
```
ZhijieZhang: One thing I forgot to mention is that users can define the UI hotkeys in their app by calling
```javascript
// see hotkeys.js
instance.hotkeys.on('space', () => {});
instance.hotkeys.off('space');
```

If we only have `keyup: true` for this key then when they define hotkeys they will see that their handler for `space` is triggered twice and this may confuse them. I'm not sure if I'm making much sense, we can discuss more about this.

I think it's actually useful to have `{ keyup: true }` here since users may want to do something on `keyup`. What do you think about this? If we have `hotkeys(key, { keyup: true }, handler);` for every key then we may need to tweak the existing logic a bit.

src/helpers/hotkeysManager.js
```diff
@@ -246,6 +255,26 @@ WebViewer(...)
...
+       */
+      space: {
+        keyup: e => {
+          e.preventDefault();
+
+          setToolModeAndGroup(store, this.prevToolName);
+          this.prevToolName = null;
+        },
+        keydown: e => {
+          e.preventDefault();
```
lbittner-pdftron: This prevents me from typing a space in a text box, for example a comment. We should make sure that no input elements are focused before triggering this. Or maybe check that the document is in focus instead

src/helpers/hotkeysManager.js
```diff
@@ -65,7 +64,17 @@ WebViewer(...)
     }
 
     // https://github.com/jaywcjlove/hotkeys#defining-shortcuts
-    hotkeys(key, handler);
+    const { keyup, keydown } = handler;
+    const keyuphandler = keyup || (() => {});
```
lbittner-pdftron: you can do this instead here:

```js
const {
    keyup = () => {},
    keydown = handler
} = handler
```

lbittner-pdftron: Also another common practice here is to define a `NOOP` function at the top of the file, like so:
```
const NOOP = () => {}
```

src/helpers/hotkeysManager.js
```diff
@@ -65,7 +64,15 @@ WebViewer(...)
     }
 
     // https://github.com/jaywcjlove/hotkeys#defining-shortcuts
-    hotkeys(key, handler);
+    const { keyup = () => {}, keydown = handler } = handler;
```
lbittner-pdftron: Can we NOOP this function please. Besides that everything looks good to me. @ZhijieZhang this ones on you now!

jiaqihuang-233: yeah sure. on it.

PR #389, opened 2019-10-29
[Feature] disable reply for annotations

jiaqihuang-233:
Need help in better variable naming.

### DIFF
--- a/src/apis/disableReplyForAnnotations.js
+++ b/src/apis/disableReplyForAnnotations.js
@@ -1,0 +1,17 @@
+/**
+ * Disable reply for annotations determined by the function passed in as parameter
+ * @param Function that takes an annotation and returns if the reply of this annotations should be disabled.
+ * @example
+ WebViewer(...)
+  .then(function(instance) {
+    /
+    instance.disableReplyForAnnotations(function(annotation) {
+      return annotation instanceof instance.Annotations.FreeHandAnnotation;
+    });
+  });
+ */
+import actions from 'actions';
+
+export default store => isReplyDisabledFunc => {
+  store.dispatch(actions.disableReplyForAnnotations(isReplyDisabledFunc));
+};
--- a/src/apis/disableReplyForAnnotations.js
+++ b/src/apis/disableReplyForAnnotations.js
@@ -2,13 +2,26 @@
  * Disable reply for annotations determined by the function passed in as parameter
  * @method WebViewer#disableReplyForAnnotations
  * @param {WebViewer~isReplyDisabled} isReplyDisabled Function that takes an annotation and returns if the reply of the annotation should be disabled.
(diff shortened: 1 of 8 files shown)


src/components/Note/Note.js
```diff
@@ -65,13 +70,12 @@ const Note = ({ annotation }) => {
   return (
     <div ref={containerRef} className={noteClass} onMouseDown={handleNoteClick}>
       <NoteContent annotation={annotation} />
-
-      <div className={repliesClass}>
+      {!isReplyDisabled && <div className={repliesClass}>
```
ZhijieZhang: Hmm, I'm not 100% sure(need to check with Matt later) but I think we still want to show existing replies. We just need to disable the textarea that's used to add a reply. Sorry for not making myself clear.

Do you think it'd be better to handle this logic in the `ReplyArea` component(just at the bottom of this file)? If you want to change that component, I think you need to merge dev because Andrey did some change to the component as well.

src/components/Note/Note.js
```diff
@@ -19,11 +19,16 @@ const propTypes = {
...
 
+const RETURN_FALSE = () => false;
+
 const Note = ({ annotation }) => {
   const { isSelected, resize } = useContext(NoteContext);
   const containerRef = useRef();
   const containerHeightRef = useRef();
 
+  const isReplyDisabledFunc = useSelector(state => selectors.getIsReplyDisabled(state)) || RETURN_FALSE;
+  const isReplyDisabled = isReplyDisabledFunc(annotation);
```
ZhijieZhang: I think you can just do something like `isReplyDisabledFunc?.(annotation)`. 
It's equivalent to
```javascript
const isReplyDisabled = isReplyDisabledFunc ? isReplyDisabledFunc(annotation) : undefined;
```
This way you don't need to have the `RETURN_FALSE` function.

src/apis/disableReplyForAnnotations.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Disable reply for annotations determined by the function passed in as parameter
+ * @param Function that takes an annotation and returns if the reply of this annotations should be disabled.
+ * @example
+ WebViewer(...)
+  .then(function(instance) {
+    /
```
ZhijieZhang: Do we need this?

src/apis/disableReplyForAnnotations.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Disable reply for annotations determined by the function passed in as parameter
+ * @param Function that takes an annotation and returns if the reply of this annotations should be disabled.
```
ZhijieZhang: I think this should be @param {function}, but can't find jsDoc documentation about this :(

PR #391, opened 2019-10-29
[Upgrade] Updated single search behaviour

andreysaf:
- Updated single search behaviour to indicate to user if there are no results


andreysaf:
image

PR #395, opened 2019-11-04
[Upgrade]: react-virtualized

jiaqihuang-233:
Remove react-list package to reduce some bundle size


src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -265,22 +267,40 @@ class ThumbnailsPanel extends React.PureComponent {
...
+              width: bounds.width,
+            });
+          }}
+        >
+          {({ measureRef }) => (
+            <div ref={measureRef} className="virtualized-thumbnails-container" >
+              <List
+                height={this.state.height}
+                width={this.state.width}
+                rowHeight={180}
```
ZhijieZhang: Let's create a constant variable for this number, as the number `180` is used twice in this component. It'd be a bit easier for us to update the height if we change it later on.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -265,22 +267,40 @@ class ThumbnailsPanel extends React.PureComponent {
...
+                rowCount={totalPages}
```
ZhijieZhang: Hmm I'm not sure if this is correct... It was `totalPages / this.state.numberOfColumns` before.
You can verify it by shrinking the browser size and see if there're multiple thumbnails in one row.

PR #397, opened 2019-11-05
[bugfix] fix worker loading after config.js is loaded

sisardor:
fix is related to this bug https://trello.com/c/s4b58qPv/550-the-preloadworker-option-and-using-setpdfworkerpath-do-not-work-together-well

The "preloadWorker" option used to load workers beforehand and setPDFWorkerPath, (and other set path functions) is used in config file. WebViewer will try to load workers first and after that process config.js file. And WV loads workers from the wrong location since setPDFWorkerPath is processes later

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -142,10 +142,6 @@ if (window.CanvasRenderingContext2D) {
       document.getElementById('app'),
     );
   });
-
-
-
-
 }
 
 window.addEventListener('hashchange', () => {
--- a/src/index.js
+++ b/src/index.js
@@ -84,9 +84,9 @@ if (window.CanvasRenderingContext2D) {
   }
 
   const { preloadWorker } = state.advanced;
-  const { PDF, OFFICE, ALL } = workerTypes;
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -108,11 +108,16 @@ if (window.CanvasRenderingContext2D) {
...
   }
 
+
   loadCustomCSS(state.advanced.customCSS);
 
   logDebugInfo(state.advanced);
 
   fullAPIReady.then(() => loadConfig()).then(() => {
+    if (preloadWorker) {
+      initTransports();
```
ZhijieZhang: Hmm why do we define a function for this? Can we move the function body into this `if` block instead?
Can you also move `const { preloadWorker } = state.advanced` here? Just to collocate better.

ZhijieZhang: Actually I think having a function is fine here since the code below are a bunch of functions. Please ignore my previous comment. I think this PR is good for merging now. @mparizeau-pdftron just need your approval.

PR #398, opened 2019-11-05
[upgrade]: Refactor PasswordModal & ToggleElementOverlay with React hooks

jiaqihuang-233:


src/components/ToggleZoomOverlay/ToggleZoomOverlay.js
```diff
@@ -0,0 +1,104 @@
...
+      setValue(zoom);
+    } else {
+      setValue(Number(e.target.value).toString());
+      zoomTo(e.target.value / 100);
+    }
+  };
+
+  return (
+    <div className="ToggleZoomOverlay">
+      <div className={['OverlayContainer', isActive ? 'active' : ''].join(' ').trim()}>
```
ZhijieZhang: Not your fault, but can we use classnames here please.

src/components/ToggleZoomOverlay/ToggleZoomOverlay.js
```diff
@@ -0,0 +1,104 @@
...
+  isActive: selectors.isElementOpen(state, 'zoomOverlay'),
+});
+
+const mapDispatchToProps = dispatch => ({
+  onClick: () => {
+    dispatch(actions.toggleElement('zoomOverlay'));
+  },
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(ToggleZoomOverlay);
```
ZhijieZhang: Same here. Let's use hook version.

src/components/ToggleZoomOverlay/ToggleZoomOverlay.js
```diff
@@ -0,0 +1,104 @@
...
+    const onZoomUpdated = () => setValue(Math.ceil(core.getZoom() * 100).toString());
+
+    core.addEventListener('documentLoaded', onDocumentLoaded);
+    core.addEventListener('zoomUpdated', onZoomUpdated);
+
+    return () => {
+      core.removeEventListener('documentLoaded', onDocumentLoaded);
+      core.removeEventListener('zoomUpdated', onZoomUpdated);
+    };
+  });
```
ZhijieZhang: I don't think we actually need to remove the listeners and then add them back on every rerender. Do you think we can just use `[]` as the second argument?

src/components/ToggleZoomOverlay/ToggleZoomOverlay.js
```diff
@@ -0,0 +1,95 @@
...
+      zoomTo(e.target.value / 100);
+    }
+  };
+
+  return (
+    <div className="ToggleZoomOverlay">
+      <div className={classNames({
+        OverlayContainer: true,
+        'active': isActive,
+      })}
```
ZhijieZhang: Hmm I feel like the styling is kind of off here. It's not a problem. I will merge the PR and take a look. Thanks!

PR #404, opened 2019-11-08
Bug fix/watermark api broken

achenPDFTron:


package-lock.json
```diff
@@ -13347,4 +13347,4 @@
       }
     }
   }
-}
+}
```
achenPDFTron: i deleted the package-lock and then commited. so i ended up creating a new one and re-copying back the content.

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -77,28 +77,26 @@ class WatermarkModal extends React.PureComponent {
...
       locationSettings,
       previousLocationSettings: locationSettings,
+      preExistingWatermark: {},
     };
     this.canvasContainerRef = React.createRef();
 
-    this.handleWatermarkRenderFxn = () => {
+    this.handleWatermarkRenderFxn = async () => {
       if (this.props.isVisible) {
+        const preExistingWatermark = await core.getWatermark();
```
achenPDFTron: store existing watermarks (if any) before we add in the watermarks created by watermark modal. when dialog closes, put back in pre-existing watermarks.

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -162,7 +160,7 @@ class WatermarkModal extends React.PureComponent {
   }
 
   // eslint-disable-next-line class-methods-use-this
-  removeAllWatermarks = () => {
+  removeWatermarkCreatedByModal = () => {
```
achenPDFTron: renamed fxn for clarity

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -77,28 +77,26 @@ class WatermarkModal extends React.PureComponent {
...
-        this.removeAllWatermarks();
+        this.removeWatermarkCreatedByModal();
+        core.setWatermark(this.state.preExistingWatermarks);
       }
     };
   }
 
-  componentDidMount() {
-    if (this.props.isVisible !== undefined) {
-      core.addEventListener('documentLoaded', this.handleWatermarkRenderFxn);
```
achenPDFTron: removed this.
from my distant memory, i wrote this to handle when watermark modal is opened on app loads.
it seems like:
- watermark modal can never be opened on app load
- it starting to look and feel weird that the watermark modal cares about the document being loaded

so i removed it for lesser complexity

src/components/PrintModal/WatermarkModal/WatermarkModal.js
ZhijieZhang: Why is `preExistingWatermark ` on state? Do you use it to rerender the view?

Also, it seems a bit weird that we define a function in the constructor. Could you tell me the reason why you are doing this?

ZhijieZhang: Hmm if we need to use it somewhere else in this component in the future then we can store it on `this` instead of `this.states`... My opinion is that properties of `this.states` should be related to how the component look like. If a variable has nothing to do with the view then we should store it on `this`, or even locally...

> I needed to define a function so that I can get its reference to that I can add/remove the event listener

I think if you define it like the rest of the functions(for example: resetForm), you can still access it via `this`?... So I think you can just define it like others. What do you think?

achenPDFTron: `My opinion is that properties of this.states should be related to how the component look like. If a variable has nothing to do with the view then we should store it on this, or even locally`
It seems I lacked a bit of understanding of `this.states` vs `this`. I cross checked what you were saying on Google as well. **I will move it out of the state and store it locally or on `this`**.

`I think if you define it like the rest of the functions(for example: resetForm), you can still access it via this?... So I think you can just define it like others`
**I'm fine with that suggestion. By doing it that way, I think it will match the consistency of how the other functions are structured.**

achenPDFTron: addressed

PR #408, opened 2019-11-13
[Feature] User bookmarks panel

khein-pdftron:
User bookmarks panel


src/components/BookmarksPanel/BookmarksPanel.js
```diff
@@ -0,0 +1,99 @@
...
+  render() {
+    const { isDisabled, display, bookmarks, addBookmark, currentPage, t } = this.props;
+
+    if (isDisabled) {
+      return null;
+    }
+
+    const pageIndexes = Object.keys(bookmarks).map(pageIndex => parseInt(pageIndex, 10));
+    return (
+      <div className="Panel bookmarks-panel" style={{ display }} data-element="bookmarksPanel">
```
ZhijieZhang: I kinda want to rename the class name to be `BookmarksPanel`, just to make it consistent with other components. For example we use `NotesPanel` instead of `notes-panel`... Let me know what you think about this.

We can probably remove this kind of class names when implementing the new UI design because customers can just select the `data-element` to customize the css styles and `data-element` is unique. What do you think?

src/components/BookmarksPanel/BookmarksPanel.js
```diff
@@ -0,0 +1,99 @@
...
+                  this.setState({ isAdding: true });
+                }}
+              >
+                {t('component.newBookmark')}
+              </div>
+            </div>
+        }
+        <div className="bookmarks-panel-row">
+          {pageIndexes.map(pageIndex => (
+            <>
```
ZhijieZhang: Can you use `React.Fragment` instead? The reason is that jsdoc can't parse `<>` when we generate UI documentations... I'm not sure if there's anyway to make this parsable for jsdoc though.

src/components/Bookmark/EditingBookmark.js
```diff
@@ -0,0 +1,50 @@
...
+          className="editing-input"
+          placeholder="Name"
+          value={this.state.bookmarkText}
+          onChange={e => {
+            this.setState({ bookmarkText: e.target.value });
+          }}
+        />
+        <div className="editing-controls">
+          <div className="editing-button editing-pad" onClick={onCancel}>
+            Cancel
```
ZhijieZhang: Need translation for  `Cancel` and `Save`. I think they are `t(action.cancel)` and `t(action.save)`.

src/redux/reducers/documentReducer.js
```diff
@@ -22,6 +22,29 @@ export default initialState => (state = initialState, action) => {
...
+      return { ...state, bookmarks: payload.bookmarks };
+    }
+    case 'ADD_BOOKMARK':
+    {
+      const newBookmarks = { ...state.bookmarks };
+      newBookmarks[payload.pageIndex] = payload.text;
+      return { ...state, bookmarks: newBookmarks };
+    }
+    case 'EDIT_BOOKMARK':
+    {
```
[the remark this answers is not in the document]
khein-pdftron: done

src/components/BookmarksPanel/BookmarksPanel.js
khein-pdftron: done

src/components/BookmarksPanel/BookmarksPanel.js
khein-pdftron: changed bookmarks-panel to bookmarksPanel

you mean by using [attribute=value] css selector with data-element?

src/components/Bookmark/EditingBookmark.js
khein-pdftron: done

src/components/BookmarksPanel/BookmarksPanel.js
ZhijieZhang: No just the className.

`className="Panel bookmarks-panel"` to `className="Panel BookmarksPanel"`

khein-pdftron: k done

PR #409, opened 2019-11-14
[Feature] Added new APIs

awy3:
For the thumbnail updated, the customer also needed these APIs.


awy3:
Changing the API to have logic in Core instead of UI

PR #410, opened 2019-11-16
[Bugfix] Fix printing issue on iOS/Android

awy3:
Safari does weird stuff when printing when printing with a small screen (can either use an iPhone or just resize the window). Trying to set the "width" to something doesn't fix it, but we can use the hack in this fix. Also when using Chrome on iOS, `document.execCommand('print')` return true even when it doesn't do anything ( need to use `window.print()`).

### DIFF
--- a/src/components/PrintModal/PrintModal.js
+++ b/src/components/PrintModal/PrintModal.js
@@ -13,6 +13,7 @@ import { getSortStrategies } from 'constants/sortStrategies';
 import actions from 'actions';
 import selectors from 'selectors';
 import LayoutMode from 'constants/layoutMode';
+import { isChromeOniOS } from 'helpers/device';
 
 import './PrintModal.scss';
 import { mapAnnotationToKey, getDataWithKey } from '../../constants/map';
@@ -393,7 +394,7 @@ class PrintModal extends React.PureComponent {
     printHandler.appendChild(fragment);
 
     // Print for Safari browser. Makes Safari 11 consistently work.
-    if (!document.execCommand('print')) {
+    if (!document.execCommand('print') || isChromeOniOS) {
       // fallback for firefox
       window.print();
     }
--- a/src/components/PrintModal/PrintModal.js
+++ b/src/components/PrintModal/PrintModal.js
@@ -13,7 +13,7 @@ import { getSortStrategies } from 'constants/sortStrategies';
 import actions from 'actions';
 import selectors from 'selectors';
 import LayoutMode from 'constants/layoutMode';
-import { isChromeOniOS } from 'helpers/device';
+import { isSafari, isChromeOniOS } from 'helpers/device';
 
 import './PrintModal.scss';
 import { mapAnnotationToKey, getDataWithKey } from '../../constants/map';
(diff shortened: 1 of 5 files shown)


src/components/PrintModal/PrintModal.js
```diff
@@ -393,7 +394,7 @@ class PrintModal extends React.PureComponent {
     printHandler.appendChild(fragment);
 
     // Print for Safari browser. Makes Safari 11 consistently work.
-    if (!document.execCommand('print')) {
+    if (!document.execCommand('print') || isChromeOniOS) {
```
ZhijieZhang: Hmm I think at first we use `window.print` for every browser and then we found we need to use `document.execCommand('print')` or Safari...

Do you think writing in this way is a bit cleaner?
```javascript
if (isSafari) {
  // Print for Safari browser. Makes Safari 11 consistently work.
  document.execCommand('print');
} else {
  window.print();
}
```

src/helpers/device.js
```diff
@@ -11,3 +11,5 @@ export const isIOS = window.navigator.userAgent.match(/(iPad|iPhone|iPod)/i) ||
 export const isAndroid = window.navigator.userAgent.match(/Android/i);
 export const isMobileDevice = isIOS || isAndroid || window.navigator.userAgent.match(/webOS|BlackBerry|IEMobile|Opera Mini/i);
 export const isMac = navigator.appVersion.indexOf('Mac') > -1;
+
+export const isChromeOniOS =  window.navigator.userAgent.match(/CriOS\/(.*?) /);
```
ZhijieZhang: One thing I'd like to discuss with you. Do you think it makes sense to use a browser detection library? For example this one https://www.npmjs.com/package/bowser

If we used this one we don't need to worry about updating this file or adding more browsers to it...

src/components/PrintModal/PrintModal.js
awy3: I can update this, but it looked like we are using  "document.execCommand('print') " as the default and fallback `to window.print()` if it fails. We are kind of changing the logic in your snippet ( but I guess if we know `document.execCommand('print')` will fail for everything beside Safari, it fine)

src/helpers/device.js
awy3: Yea it seems like a good idea to this rather than doing it ourselves

ZhijieZhang: kk sounds good I will take a closer look at this package and talk to Matt later :)

src/components/PrintModal/PrintModal.js
ZhijieZhang: Yeah my snippet changed the logic but since at the very beginning we only do `window.print()`, I think we can just use this for all the browsers other than Safari. Not sure if I'm making sense.

awy3:
I added a fix for Android because I was noticing printing in Chrome on Android was broken. It looks like it broke after 13e0ae0 (this work for desktop but we get `DOMException: Blocked a frame with origin "http://192.168.10.203:3000" from accessing a cross-origin frame` error  in Android)

src/components/PrintModal/PrintModal.js
awy3: I updated the code

PR #413, opened 2019-11-18
[Docs] Updates UI Docs to use proper types

lbittner-pdftron:


src/apis/setHeaderItems.js
```diff
@@ -74,7 +75,7 @@ export default store => callback => {
 /**
  * A class which contains header APIs.<br/><br/>
  * <span style="color: red; font-size: 1.2em; font-weight: bold">⚠</span> You must NOT instantiate this yourself. Access the header instance in {@link WebViewer#setHeaderItems setHeaderItems} as follows:
- * @name WebViewer.Header
+ * @name Header
```
ZhijieZhang: Hmm if we do this to `Header` as well, then what's the way to find the documentation for it? Currently it's here https://www.pdftron.com/api/web/WebViewer.html#toc2__anchor.

lbittner-pdftron: It will be in the "classes" menu and will be linked to from appropriate methods

lbittner-pdftron:
I had to remove the `~` syntax and replace with dot notation :'(

PR #415, opened 2019-11-19
[Bugfix] Redraw signature canvas on resize or rotate

awy3:
For a ticket a  customer was saying the signature canvas clear when the screen rotate (I'm guessing they are on a mobile device), so I made it redraw after it get clear.  

signatureResize

Another solution would be do something like
```
        const image = new Image();
        const ctx = canvas.getContext('2d');
        image.onload = function() {
          ctx.drawImage(image, 0, 0);
        };
        image.src = canvas.toDataURL();
```

This might be good if we don't want to use "signatureTool.drawAnnot"

### DIFF
--- a/src/components/SignatureModal/SignatureModal.js
+++ b/src/components/SignatureModal/SignatureModal.js
@@ -38,8 +38,8 @@ class SignatureModal extends React.PureComponent {
 
   componentDidMount() {
     this.setUpSignatureCanvas();
-    window.addEventListener('resize', this.setSignatureCanvasSize);
-    window.addEventListener('orientationchange', this.setSignatureCanvasSize);
+    window.addEventListener('resize', this.onResize);
+    window.addEventListener('orientationchange', this.onRotate);
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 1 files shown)


src/components/SignatureModal/SignatureModal.js
```diff
@@ -100,6 +100,11 @@ class SignatureModal extends React.PureComponent {
     const { width, height } = canvas.getBoundingClientRect();
     canvas.width = width;
     canvas.height = height;
+
+    if (!this.signatureTool.isEmptySignature()) {
+      // if the canvas is resized the content get cleared. So redraw it
+      this.signatureTool.drawAnnot();
```
ZhijieZhang: Hmm `drawAnnot` is removed in 6.0 so I think we should go with the other way you suggested.

Also I think it'd be better if we have a separate function for this. Redrawing the canvas in a function called `setSignatureCanvasSize` is kind of weird imo.

Do you think it makes more sense to just clear the signature when an orientation happens? The signature that was drawn is out of position anyways because of the canvas size change?

ZhijieZhang: And I think for redraw we can use `ctx.getImageData` and `ctx.putImageData `

awy3: Rotating and resizing isn't perfect but isn't too bad 

signatureResize2

awy3: I refactor the code, I didn't end up with "getImageData" and "putImageData" because I wanted to do "drawImage(imageData, 0, 0, newWidth, newHeight)" for resizing the image

PR #416, opened 2019-11-20
[Docs] Added documentation for bookmarks api

khein-pdftron:


src/apis/exportBookmarks.js
```diff
@@ -1,3 +1,22 @@
...
+    // Save the annotation data for doc123
+    const bookmarks = instance.exportBookmarks();
+    const bookmarksString = JSON.stringify(bookmarks);
+    fetch('/server/bookmarksHandler.js?documentId=doc123', {
+      method: 'POST',
+      body: bookmarksString // written into a json file on server
+    });
+  });
+ */
+/**
```
ZhijieZhang: I think this should just be `*`, otherwise the `returns` won't be included in the documentation.

Screen Shot 2019-11-22 at 12 30 04 PM

khein-pdftron: Fixed now.

PR #417, opened 2019-11-21
[Feature] Thumbnail reordering and merging

awy3:
This is the update for reordering and merging using the thumbnail panel. It start disabled but you can use `readerControl.enableFeatures(['ThumbnailMerging'])` and `readerControl.enableFeatures(['ThumbnailReordering'])` to get it to work

This depends on "readerControl.docViewer.getDocument().mergeDocument()", which is in another pull request for 6.0.


src/components/Thumbnail/Thumbnail.js
```diff
@@ -15,17 +15,23 @@ class Thumbnail extends React.PureComponent {
...
     pageLabels: PropTypes.array.isRequired,
     canLoad: PropTypes.bool.isRequired,
+    isSelected: PropTypes.bool,
     onLoad: PropTypes.func.isRequired,
     onCancel: PropTypes.func.isRequired,
     onRemove: PropTypes.func.isRequired,
     updateAnnotations: PropTypes.func,
     closeElement: PropTypes.func.isRequired,
+    onDragStartCallback: PropTypes.func,
+    onDragOverCallback: PropTypes.func,
```
ZhijieZhang: I prefer `onDragStart` and `onDragOver`. I think it's common that variables start with `on` are callbacks, so we can probably skip the `callback` part? :)

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -271,6 +378,8 @@ class ThumbnailsPanel extends React.PureComponent {
         className="Panel ThumbnailsPanel"
         style={{ display }}
         data-element="thumbnailsPanel"
+        ref={this.thumbnailsPanel}
+        onDragOver={this.dragOverHandler}
```
ZhijieZhang: I think we can just use `e => e.preventDefault` here instead of having a function. What do you think? We do similar things in other components.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -38,6 +50,10 @@ class ThumbnailsPanel extends React.PureComponent {
     core.addEventListener('pageNumberUpdated', this.onPageNumberUpdated);
     core.addEventListener('annotationHidden', this.onAnnotationChanged);
     window.addEventListener('resize', this.onWindowResize);
+
+    if (this.thumbnailsPanel && this.thumbnailsPanel.current) {
+      this.thumbnailsPanel.current.addEventListener('drop', this.onDrop);
```
ZhijieZhang: Hmm maybe I'm missing something, but can we just pass an onDrop prop to the thumbnails panel div? If we can do that we don't need to ref it.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -55,6 +71,75 @@ class ThumbnailsPanel extends React.PureComponent {
...
+      }
+
+      core.movePages([currentPage], targetPageNumber);
+    }
+
+    this.setState({ draggingOverPageIndex: null });
+  }
+
+  onPageComplete = () => {
+    if (this.afterMovePageNumber) {
```
ZhijieZhang: Hmm this one isn't called anywhere in the component. Just to confirm we need this in the later commits?

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -55,6 +71,75 @@ class ThumbnailsPanel extends React.PureComponent {
...
+
+  onDragOver = (e, index) => {
+    // prevent opening pdf dropped in
+    e.preventDefault();
+    const thumbnail = e.target.getBoundingClientRect();
+
+    if (e.pageY > (thumbnail.y + thumbnail.height / 2)) {
+      this.setState({ draggingOverPageIndex: index, isDraggingOverTopHalf: false });
+    } else {
+      this.setState({ draggingOverPageIndex: index, isDraggingOverTopHalf: true });
```
ZhijieZhang: Hmm the other way of writing this is
```javascript
this.setState({
  draggingOverPageIndex: index,
  isDraggingOverTopHalf: e.pageY > (thumbnail.y + thumbnail.height / 2)
})
```
This doesn't use ifelse, but I don't mind the current way.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -55,6 +71,75 @@ class ThumbnailsPanel extends React.PureComponent {
...
+    e.preventDefault();
+    const { isThumbnailMerging } = this.props;
+    const { draggingOverPageIndex, isDraggingOverTopHalf } = this.state;
+    const { files } = e.dataTransfer;
+
+    if (isThumbnailMerging && files.length) {
+      const file = files[0];
+      const insertTo = isDraggingOverTopHalf ? draggingOverPageIndex + 1 : draggingOverPageIndex + 2;
+
+      window.readerControl.mergeDocument(file, insertTo).then(() => {
```
ZhijieZhang: Hmm in general we avoid to call readerControl within our components. The reason is that we want to make clear that functions of readerControl are for external uses.

In this case what we do is to create a new file under the helpers folder, and use that here. And in the apis folder we import the helper function to "expose" it... I know it's essentially doing the same thing, but I think we just want to remain the same mental model. Sorry

src/apis/mergeDocument.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Merge a file into the currently opened document
+ * @method WebViewer#mergeDocument
+ * @param {(string|File)} documentPath Path to the document OR <a href='https://developer.mozilla.org/en-US/docs/Web/API/File' target='_blank'>File object</a> if opening local file.
+ * @param {number} position Position to put the merged document, default to end of file if nothing entered
+ * @returns {Promise<any>} a promise that resolves on completion
+ * @example // 5.1 and after
```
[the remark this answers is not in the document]
awy3: Thanks for pointing this out, I'll copy the jsDoc from the core Document.js file

src/apis/mergeDocument.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Merge a file into the currently opened document
+ * @method WebViewer#mergeDocument
+ * @param {(string|File)} documentPath Path to the document OR <a href='https://developer.mozilla.org/en-US/docs/Web/API/File' target='_blank'>File object</a> if opening local file.
```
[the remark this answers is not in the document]
awy3: Thanks for pointing this out, I'll copy the jsDoc from the core Document.js file

src/apis/mergeDocument.js
```diff
@@ -0,0 +1,25 @@
...
+ * @example // 5.1 and after
+WebViewer(...)
+  .then(function(instance) {
+    instance.mergeDocument('https://www.pdftron.com/downloads/pl/test.pdf', 1);
+  });
+ */
+import actions from 'actions';
+
+export default store => (documentToMerge, position) => {
+  store.dispatch(actions.openElement('progressModal'));
```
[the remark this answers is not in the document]
awy3: I'll update it to "loadingModal", before when I was taking care of the partRetriever stuff in the UI, I had more control  and could show progress ( I think I can still do that but not sure if it worth it since we are doing other UI changes soon)

src/apis/mergeDocument.js
```diff
@@ -0,0 +1,25 @@
...
+ * @returns {Promise<any>} a promise that resolves on completion
+ * @example // 5.1 and after
+WebViewer(...)
+  .then(function(instance) {
+    instance.mergeDocument('https://www.pdftron.com/downloads/pl/test.pdf', 1);
+  });
+ */
+import actions from 'actions';
+
+export default store => (documentToMerge, position) => {
```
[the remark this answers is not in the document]
awy3: In the ThumbnailPanel, I need it to allow users to drag and drop file to merge.

I can provide you with more information on customer requirements  (not sure if I should post it here since it's public)

src/components/Thumbnail/Thumbnail.js
```diff
@@ -89,14 +95,24 @@ class Thumbnail extends React.PureComponent {
...
   render() {
-    const { index, currentPage, pageLabels } = this.props;
+    const { index, currentPage, pageLabels, isSelected, draggable } = this.props;
     const isActive = currentPage === index + 1;
     const pageLabel = pageLabels[index];
 
     return (
-      <div className={`Thumbnail ${isActive ? 'active' : ''}`}>
-        <div className="container" ref={this.thumbContainer} onClick={this.handleClick}></div>
+      <div className={`Thumbnail ${isActive ? 'active' : ''} ${isSelected ? 'selected' : ''}`} onDragOver={this.onDragOverHandler}>
```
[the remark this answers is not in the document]
awy3: I'll remove it for now, I'll need it later when we add selecting thumbnails

src/components/Thumbnail/Thumbnail.js
```diff
@@ -15,17 +15,23 @@ class Thumbnail extends React.PureComponent {
...
+    onDragOverCallback: PropTypes.func,
+    draggable: PropTypes.bool,
   }
 
   constructor(props) {
     super(props);
     this.thumbContainer = React.createRef();
     this.onLayoutChangedHandler = this.onLayoutChanged.bind(this);
+    this.onDragStartHandler = this.onDragStart.bind(this);
+    this.onDragOverHandler = this.onDragOver.bind(this);
```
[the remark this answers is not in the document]
awy3: your right, binding them wasn't needed in this case

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: We can do that but I normally avoid creating anonymous function in the render but if we are doing it in other places I can change this to just

```
onDragOver={e => e.preventDefault()}
```

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: I updated it to use the "onDrop" prop, I can't remember why I was doing it like this

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: Thank you for pointing this out, I forgot to add
```
    core.addEventListener('pageComplete', this.onPageComplete);
```

This is so if you are reordering pages, it'll go to the last page you were on. So if you are on page 1 and move it to page 5,  it'll go to page 5 (previously page 1) after it moved.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: Thank you for pointing this out, I actually like your ternary statement more. I'll update it to that

src/components/Thumbnail/Thumbnail.js
awy3: I have renamed these

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: I actually ended getting rid of this, in the "Thumbnail" "onDragOver" callback, I added 
`e.stopPropagation();` so it never get throw dragOver event for the parent.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: I didn't end up adding a helper but added it to `core`

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -28,6 +32,9 @@ class ThumbnailsPanel extends React.PureComponent {
       canLoad: true,
       height: 0,
       width: 0,
+      draggingOverPageIndex: null,
+      isDraggingOverTopHalf: false,
+      afterMovePageNumber: null,
```
ZhijieZhang: why is this on state? Does this variable related to how the thumbnail panel looks like?

src/core/mergeDocument.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * https://www.pdftron.com/api/web/CoreControls.Document.html#mergeDocument__anchor
+ */
+import actions from 'actions';
+
+export default (dispatch, documentToMerge, position) => {
+  dispatch(actions.openElement('loadingModal'));
+  return window.readerControl.docViewer.getDocument().mergeDocument(documentToMerge, position)
```
ZhijieZhang: Hmm as I said we don't use readerControl internally. APIs on the readerControl are only for external usage. Let me know if you are not clear about what I mean.

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: I moved the "afterMovingPageNumber" out of the state. It not used for appearance so it shouldn't be there

src/core/mergeDocument.js
awy3: Thanks for pointing this out, I added the API file but forgot to get rid of the "readerControl", now this uses `window.docViewer`

PR #420, opened 2019-11-23
[Feature] Delayed & faded in progress modal

sisardor:
The feature delays progress modal for 2 seconds so that UI will wait and fade-in the progress modal if the file downloading is too slow.

### DIFF
--- a/src/helpers/loadDocument.js
+++ b/src/helpers/loadDocument.js
@@ -29,12 +29,7 @@ export default (state, dispatch, extraOptions) => {
             partRetriever.setErrorCallback(fireError);
           }
 
-          // Delaying progress modal for 2 sec, so that UI will wait and
-          // fade-in the progress modal if the file downloading is too slow
-          setTimeout(() => {
-            dispatch(actions.openElement('progressModal'));
-          }, 2000);
-
+          dispatch(actions.openElement('progressModal'));
           core.loadAsync(partRetriever, { ...extraOptions, ...docOptions });
         })
         .catch(error => {
--- a/src/helpers/loadDocument.js
+++ b/src/helpers/loadDocument.js
@@ -29,10 +29,11 @@ export default (state, dispatch, extraOptions) => {
             partRetriever.setErrorCallback(fireError);
           }
 
-          // Delay opening of progress for 1 sec
+          // Delaying progress modal for 2 sec, so that UI will wait and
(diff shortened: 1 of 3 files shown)


sisardor:
Delaying time was decided to be 2 seconds.

src/helpers/loadDocument.js
```diff
@@ -29,7 +29,12 @@ export default (state, dispatch, extraOptions) => {
             partRetriever.setErrorCallback(fireError);
           }
 
-          dispatch(actions.openElement('progressModal'));
+          // Delaying progress modal for 2 sec, so that UI will wait and
+          // fade-in the progress modal if the file downloading is too slow
+          setTimeout(() => {
+            dispatch(actions.openElement('progressModal'));
+          }, 2000);
```
ZhijieZhang: Can we just use css transition or animation delay of 2s instead of using setTimeout?
I feel like a pure css solution is better. Something like
```
  &.open {
    opacity: 1;
    transition: opacity 2s 1s;
  }

  &.closed {
    opacity: 0;
  }
```
The transition of opacity will wait for 2 secs before starting.
If we do this we also don't need the extra logic in openElements, because if the document is loaded or if there's an error loading a document, the progress modal will be closed immediately.

src/redux/actions/exposedActions.js
```diff
@@ -29,7 +29,18 @@ export const openElement = dataElement => (dispatch, getState) => {
     }
     dispatch(setActiveLeftPanel(dataElement));
   } else {
-    dispatch({ type: 'OPEN_ELEMENT', payload: { dataElement } });
+    if (dataElement === 'progressModal') {
+      // Opening progress modal will not be necessary if the document
```
ZhijieZhang: See the other comment, I don't think this is necessary...

src/helpers/loadDocument.js
sisardor: Yeah, sure we can do that,  I'll revert add CSS

ZhijieZhang:
Thanks. The code part looks good to me. I just want to make sure that if you have talked to the design team about this? Just want to make sure that everyone is on the same page...

sisardor:
I think in the last sprint meeting Ivan has talked about this delaying progress bar for fast loading PDF files and the design team knows about it.  But I let them know about it!

PR #427, opened 2019-11-28
[Update] Move the notes panel outside of the left panel and move it to the right

ZhijieZhang:


src/components/NotesPanel/NotesPanel.js
```diff
@@ -229,16 +230,14 @@ const NotesPanel = ({ display }) => {
...
   }
 
-  // when either of the other two panel tabs is clicked, the "display" prop will become "none"
-  // like other two panels, we should set the display style of the div to props.display
-  // but if we do this, sometimes a maximum updates errors will be thrown from react-virtualized if we click the other two panels
-  // it looks like the issue is reported here: https://github.com/bvaughn/react-virtualized/issues/1375
-  // the PR for fixing this issue has been merged but not yet released so as a workaround we are unmounting
-  // the whole component when props.display === 'none'
-  // this should be changed back after the fixed is released in the next version of react-virtualized
-  return isDisabled || display === 'none' ? null : (
```
ZhijieZhang: Actually I found that we don't need to unmount the notes panel in this case because it's not part of the left panel anymore, which means it won't receive a display `none` prop :)

src/redux/actions/exposedActions.js
```diff
@@ -1,5 +1,5 @@
 import core from 'core';
-import isDataElementPanel from 'helpers/isDataElementPanel';
+import isDataElementLeftPanel from 'helpers/isDataElementLeftPanel';
```
ZhijieZhang: I changed the name to include `Left`, though the new name is kind of weird... I think we can keep this for now? haha

PR #428, opened 2019-11-29
[Feature]thumbnail page controls

awy3:
Use "readerControl.enableElement('thumbnailControl')" to make the control appear

image


src/components/ThumbnailsPanel/ThumbnailsPanel.scss
```diff
@@ -18,4 +18,15 @@
     border-color: $gray87;
     margin: 0px;
   }
+
+  .columnsOfThumbnails{
```
ZhijieZhang: Can you add one space before the `{`?

src/constants/pageRotation.js
```diff
@@ -0,0 +1,6 @@
+export const PageRotation = {
+  e_0: window.CoreControls.PageRotation.e_90,
+  e_90: window.CoreControls.PageRotation.e_90,
+  e_180: window.CoreControls.PageRotation.e_180,
+  e_270: window.CoreControls.PageRotation.e_270,
```
ZhijieZhang: Hmm... I feel like we can just do `window.CoreControls.PageRotation` in our components. No need to wrap another "constant layer" over it?

src/helpers/removePages.js
```diff
@@ -0,0 +1,27 @@
+import core from 'core';
+import actions from 'actions';
+import i18next from 'i18next';
+
+export default pages => dispatch => {
```
ZhijieZhang: I think we can just move the logic into the thumbnail control component. I don't think we are reusing this function right?

src/redux/initialState.js
```diff
@@ -13,7 +13,12 @@ import actions from 'actions';
 
 export default {
   viewer: {
-    disabledElements: {},
+    disabledElements: {
+      thumbnailControl: {
```
ZhijieZhang: Can you move the logic to `setDefaultDisabledElements`? It introduces indirection but the reason why we did that is sometimes we need to conditionally disable an element based on the constructor options.

src/components/WarningModal/WarningModal.js
```diff
@@ -25,15 +25,18 @@ class WarningModal extends React.PureComponent {
...
   componentDidUpdate(prevProps) {
-    if (!prevProps.isOpen && this.props.isOpen) {
+    const { keepOpen, isOpen, closeElements } = this.props;
+
+    if (!prevProps.isOpen && isOpen) {
       // TODO want to do make something like'closeOtherElements', but that for latter
-      const popUpToClose = getPopupElements().filter(
-        ele => ele !== 'warningModal',
+      const popupElementsToClose = getPopupElements().filter(
+        ele => ele !== 'warningModal' && (!keepOpen || !keepOpen.includes(ele)),
```
ZhijieZhang: Can we tell me why we wanted to close the left panel when the warning modal is showing up? I feel like we can just keep the panels open... If we do that we don't need this extra logic.

awy3: There was no reason for us to have it,  want to keep the left panel open when the warning (and error, which already doing this ) modal come up.

I update the `componentDidUpdate` function to filter out left panel and removed the "keepOpen" logic. I was thinking that logic would be good for later but we'll see if "later" ever happens

src/redux/initialState.js
awy3: I moved the logic to `setDefaultDisabledElements`

src/helpers/removePages.js
awy3: I removed the helper and moved all the logic to the thumbnail panel

src/constants/pageRotation.js
awy3: I have removed this constant

src/components/ThumbnailsPanel/ThumbnailsPanel.scss
awy3: I fixed the spacing to match other scss files

src/components/ThumbnailControls/ThumbnailControls.scss
```diff
@@ -0,0 +1,24 @@
...
+        display: inline-block;
+        height: 24px;
+        padding: 0px;
+        width: 24px;
+        min-width: 24px;
+        margin: 0px 8px;
+
+
+        &:hover {
+            background-color: var(--icon-color);
```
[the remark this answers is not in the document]
awy3: Thanks for pointing this out. I updated the css and did some basic IE testing

src/components/ThumbnailControls/ThumbnailControls.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../constants/panel';
+
+.thumbnailControls {
+    text-align: center;
+    z-index: 2;
+
+    .Button{
```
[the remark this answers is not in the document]
awy3: This has been fixed

src/components/Thumbnail/Thumbnail.js
```diff
@@ -92,6 +98,23 @@ class Thumbnail extends React.PureComponent {
     }
   }
 
+  handleDelete = () => {
+    const { index, removePages, showWarningMessage } = this.props;
+
+    const message = i18next.t('option.thumbnailPanel.deleteWarningMessage');
+    const title = i18next.t('option.thumbnailPanel.deleteWarningTitle');
+    const confirmBtnText = i18next.t('option.thumbnailPanel.deleteWarningConfirmText');
```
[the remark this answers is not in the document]
awy3: I updated it to use `withTranslation`

src/components/Thumbnail/Thumbnail.js
```diff
@@ -123,6 +147,8 @@ const mapStateToProps = state => ({
 
 const mapDispatchToProps = {
   closeElement: actions.closeElement,
+  removePages: pageNumbers => removePages(pageNumbers),
```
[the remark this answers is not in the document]
awy3: I removed the helper and now we are calling `core.removePages` instead of adding this to the props

i18n/translation-en.json
```diff
@@ -197,6 +197,15 @@
       "text": "Text",
       "colors": "Colors",
       "resetAllSettings": "Reset All Settings"
+    },
+    "thumbnailPanel": {
+      "deleteWarningTitle": "Delete Page",
+      "deleteWarningMessage": "Are you sure you want to delete the selected page(s). This can't be undone",
+      "deleteWarningConfirmText": "OK",
```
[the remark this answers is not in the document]
awy3: I can remove `Ok` but for delete I can't use `action.deletel` or else I get something like
image

Is there a way to get rid of the hot key part of the text?

src/components/Thumbnail/Thumbnail.js
```diff
@@ -103,14 +126,15 @@ class Thumbnail extends React.PureComponent {
...
-    const { index, currentPage, pageLabels, isDraggable } = this.props;
+    const { index, currentPage, pageLabels, isDraggable, isThumbnailControlDisabled } = this.props;
     const isActive = currentPage === index + 1;
     const pageLabel = pageLabels[index];
 
     return (
       <div className={`Thumbnail ${isActive ? 'active' : ''} `} onDragOver={this.onDragOver}>
         <div className="container" ref={this.thumbContainer} onClick={this.handleClick} onDragStart={this.onDragStart} draggable={isDraggable}></div>
         <div className="page-label">{pageLabel}</div>
+        {isActive && !isThumbnailControlDisabled && <ThumbnailControls index={index} data-element="thumbnailControl" handleDelete={this.handleDelete} />}
```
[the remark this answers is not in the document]
awy3: I moved this logic into `ThumbnailControls` and refactor `ThumbnailControls` to use hook

src/components/Thumbnail/Thumbnail.js
```diff
@@ -92,6 +98,23 @@ class Thumbnail extends React.PureComponent {
     }
   }
 
+  handleDelete = () => {
```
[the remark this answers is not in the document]
awy3: I moved the delete logic into `ThumbnailControls`

i18n/translation-en.json
```diff
@@ -197,6 +197,15 @@
       "text": "Text",
       "colors": "Colors",
       "resetAllSettings": "Reset All Settings"
+    },
+    "thumbnailPanel": {
+      "deleteWarningTitle": "Delete Page",
+      "deleteWarningMessage": "Are you sure you want to delete the selected page(s). This can't be undone",
```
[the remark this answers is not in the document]
awy3: So something like
```
  "warning": {
    "deletePage" :{
      "deleteTitle": "Delete Page",
      "deleteMessage": "Are you sure you want to delete the selected page(s). This can't be undone",
      "deleteLastPageMessage": "This would result in no pages remaining. Please leave at least one page",
    },
    "redaction":{
      "applyTile": "Apply Redaction",
      "applyMessage": "This action will permanently remove all items selected for redaction. It cannot be undone.",
    }
  },
```
And we call it with `warning.deletePage.deleteMessage`. Or were you thinking it would be `option.warningModel.deleteMessage` or `option.warningModal.redactionApplyMessage`

src/components/Thumbnail/Thumbnail.js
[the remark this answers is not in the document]
awy3: I actually remove this later since the warning message get generated in the `ThumbnailControls` now

i18n/translation-en.json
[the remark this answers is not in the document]
ZhijieZhang: Yeah I think either of them is fine. Do you think it make sense?

src/components/Thumbnail/Thumbnail.js
```diff
@@ -66,8 +67,11 @@ class Thumbnail extends React.PureComponent {
     const didPageChange = contentChanged.some(changedPage => currentPageStr === changedPage);
     const didPageMove = Object.keys(moved).some(movedPage => currentPageStr === movedPage);
     const isPageRemoved = removed.indexOf(currentPage) > -1;
+    const newPageCount = pageLabels.length - removed.length;
 
-    if (isPageAdded || didPageChange || didPageMove || isPageRemoved) {
+    if (removed.length > 0 && index + 1 > newPageCount) {
+      // don't load thumbnail if it's going to be removed
```
ZhijieZhang: Hmm... Can we use `return` instead?

src/components/WarningModal/WarningModal.js
```diff
@@ -28,12 +28,14 @@ class WarningModal extends React.PureComponent {
...
   componentDidUpdate(prevProps) {
-    if (!prevProps.isOpen && this.props.isOpen) {
+    const { isOpen, closeElements } = this.props;
+
+    if (!prevProps.isOpen && isOpen) {
       // TODO want to do make something like'closeOtherElements', but that for latter
-      const popUpToClose = getPopupElements().filter(
-        ele => ele !== 'warningModal',
+      const popupElementsToClose = getPopupElements().filter(
+        ele => ele !== 'warningModal' && ele !== 'leftPanel',
```
ZhijieZhang: Do you think it's better if just we just remove `leftPanel` in `getPopupElements`? Also warningModal isn't in the popup elements so we don't need to filter it out I think?

src/components/Thumbnail/Thumbnail.js
awy3: I updated this

i18n/translation-en.json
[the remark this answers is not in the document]
awy3: I ended up moving the message to the top level so it's now `warning.deletePage.deleteMessage`

src/components/WarningModal/WarningModal.js
awy3: Right now it seems the `warningModal` is the only thing that is using `getPopupElements` so I think now would be a good time to discuss how it should be used. 

Lot of compounds are just hard coding stuff to close like MenuOverlay.js
https://github.com/PDFTron/webviewer-ui/blob/7c1be32d0b448980e1785cb6d45c73bd97f86933/src/components/MenuOverlay/MenuOverlay.js#L47

and SearchOverlay.js
https://github.com/PDFTron/webviewer-ui/blob/7c1be32d0b448980e1785cb6d45c73bd97f86933/src/components/SearchOverlay/SearchOverlay.js#L85

The idea was to have a list of all "popup" somewhere so if a new one was added we didn't need to update 5+ components. I will remove "leftPanel" from the list because I don't think we want to close it in most cases.

Also after this PR, it would be a good time to have other components start using the `getPopupElements` helper (or I should remove it if it doesn't seem helpful or refactor it into a "closeOtherPopupElement" helper)

PR #431, opened 2019-12-03
Add Wildcard option to advanced search.

bollain:
QA Steps:
- Launch WebViewer
- In the search panel, open it to see advanced search modes. 
- There should now be a Wildcard option
image

- Do a few wildcard searches to verify it works.


bollain:
@ZhijieZhang there is a linter failure for an unused constant - not sure if it is OK to remove it.

ZhijieZhang:
> @ZhijieZhang there is a linter failure for an unused constant - not sure if it is OK to remove it.

It's removed in dev branch so if you merge dev I think it should go away :)

src/components/SearchOverlay/SearchOverlay.js
```diff
@@ -357,6 +365,9 @@ class SearchOverlay extends React.PureComponent {
             <Input id="case-sensitive-option" type="checkbox" ref={this.caseSensitiveInput} onChange={this.onChangeCaseSensitive} label={t('option.searchPanel.caseSensitive')} />
             <Input id="whole-word-option" type="checkbox" ref={this.wholeWordInput} onChange={this.onChangeWholeWord} label={t('option.searchPanel.wholeWordOnly')} />
           </div>
+          <div className={`wildcard ${isSearchPanelOpen ? 'visible' : ''}`}>
+              <Input id="wild-card-option" type="checkbox" ref={this.wildcardInput} onChange={this.onChangeWildcard} label={t('option.searchPanel.wildcard')} />
```
ZhijieZhang: Hmm it'd be nice if we can move it to the div above.
Could you try playing around the `<div className="options ...">` div to see if you can use flex-wrap to put these three checkboxes into one div?

ZhijieZhang: Also the indentation should be 2 instead of 4.

ZhijieZhang: Looks good! One more thing, can you also add a line below 83 so that when we programmatically search the text with wildcard option being true, the checkbox will be checked?

You can verify if it's working by:
```javascript
instance.searchTextFull('You', { wildcard: true })
```

or switch to the iframe using the dev tool and then in the console
```javascript
readerControl.searchTextFull('You', { wildcard: true })
```

bollain: Made changes as requested, thanks!

src/components/SearchOverlay/SearchOverlay.scss
```diff
@@ -147,10 +147,25 @@
...
         padding: 5px 0 0 20px;
       }
     }
 
+    .search-option {
+      width: 115px;
+    }
+
+    .wildcard {
+      display: none;
```
ZhijieZhang: Sorry I'm having a little trouble finding where this classname is used... Can you tell me which element is using this?

bollain: Good catch, this was the class I was using when I had the wildcard option in its own div. Removed!

PR #432, opened 2019-12-03
Feature/watermark modal enhancement

achenPDFTron:


src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -37,14 +38,22 @@ const FORM_FIELD_KEYS = {
...
+  font: 'font',
+  isBolded: 'isBolded',
+  isItalic: 'isItalic',
+  isUnderlined: 'isUnderlined',
 };
 
 const DEFAULT_VALS = {
   [FORM_FIELD_KEYS.location]: WATERMARK_LOCATIONS.CENTER,
-  [FORM_FIELD_KEYS.fontSize]: FONT_SIZES[FONT_SIZES.length / 2],
+  [FORM_FIELD_KEYS.fontSize]: FONT_SIZES[FONT_SIZES.length - 1],
```
achenPDFTron: changed it as the default font size was a bit too small

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -310,7 +327,18 @@ class WatermarkModal extends React.PureComponent {
 
               </div>
               <div className="form-field">
-
+                <label>
+                  {t(`option.watermark.font`)}
+                </label>
+                <select
```
ZhijieZhang: If an element has `data-element` then users should be able to disable it by using `instance.disableElements`. You can either hook up with redux or just remove `data-element` if you don't want users to be able to disable it.

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -337,14 +364,37 @@ class WatermarkModal extends React.PureComponent {
...
                 </label>
-                <div
-                  data-element="currentColorCell"
-                  className="cell"
-                  style={{ backgroundColor: formInfo[FORM_FIELD_KEYS.color].toHexString() }}
-                  onClick={() => this.setColorPaletteVisibility(!this.state.isColorPaletteVisible)}
-                >
+                <div className="style-container">
+                  <div
+                    data-element="currentColorCell"
```
ZhijieZhang: Same here

src/components/PrintModal/WatermarkModal/WatermarkModal.js
```diff
@@ -337,14 +364,37 @@ class WatermarkModal extends React.PureComponent {
...
+                    className="Button"
```
ZhijieZhang: Hmm do you think it make sense that we change the Button component so that it's possible to just render a div with nothing inside? If we do that you can just use a `<Button />` here

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: I don't think I want users to be able to disable it.
However, I noticed that the end to end tests I wrote a while back look for some of these data elements so that I can fill out the form properly.

What would be your suggestion for this problem?
I'm leaning towards keeping data element? I think the user should know what he/she is doing is he/she truly wants to disable these watermark form elements?

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: I believe we are already doing that .... ?
image

For my case, content is undefined, so it ends up being a div with nothing inside...?

As for a button component that just contains a div ... ? I don't think I have seen that before and it sounds a little bit weird?

I think if the concern here is this not being a button component when it should be, I think we can just add an extra prop called styles which is of type CSS style?

src/components/PrintModal/WatermarkModal/WatermarkModal.js
ZhijieZhang: If you don't want users to be able to disable it then you need to remove `data-element`. Otherwise it may confuse them if they are following this https://www.pdftron.com/documentation/web/guides/hiding-elements guide.

I think for now you can give it an id or a class name and change cypress to use these attributes. Just to make sure that we are on the same page, when I say users I mean developers, not end users...

src/components/PrintModal/WatermarkModal/WatermarkModal.js
ZhijieZhang: Hmmm sorry, didn't know that we are already doing this... I think it's ok to just use a Button component here. I'm not very sure what you mean by `I think we can just add an extra prop called styles which is of type CSS style`. Perhaps you can elaborate a bit more?

achenPDFTron: Here is what I mean,
1. Create a new prop in the Button component called styles;
2. use that prop inside the button component

Example usuage:
``` 
<button
 styles={ {color: 'blue'} }
/>

button.jsx
<Button style={styles}></Button>
```

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: done. for some elements that are react components, i left them as "dataElements"

src/components/PrintModal/WatermarkModal/WatermarkModal.js
achenPDFTron: see above.

ZhijieZhang:
Hi Anthony, when you have time can you change this PR's target branch to be master? I think you need to close this and open another one.

achenPDFTron:
@ZhijieZhang i have changed it to be merged into master

src/components/PrintModal/WatermarkModal/WatermarkModal.js
ZhijieZhang: Sorry I missed your reply.

I see what you mean now. But if you chose to use a `Button` component here, you don't need to use the `style` prop because the `Button` component has the `Button` class and you are styling it in the scss file(https://github.com/PDFTron/webviewer-ui/pull/432/files#diff-43b4e2cfdcecbbc6852c6706e5302a27R113).

I think we generally want to avoid using inline css because a customer would need to use `!important` to override it if he's using the custom `css` option. Anyways I think this PR is good to be merged in without this change. Currently just waiting for https://github.com/XodoDocs/webviewer/pull/495 to be merged.

PR #438, opened 2019-12-14
Bug fix/accessible mode page push down

achenPDFTron:
bug fix for https://trello.com/c/3aQH02v2/635-in-accessiblemode-sometimes-the-page-gets-pushed-down-by-the-text-element

### DIFF
--- a/src/event-listeners/onPageComplete.js
+++ b/src/event-listeners/onPageComplete.js
@@ -6,6 +6,7 @@ export default store => pageIndex => {
     core.getDocument().loadPageText(pageIndex, text => {
       const textContainer = document.createElement('div');
       const className = 'accessible-text';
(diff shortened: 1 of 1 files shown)


src/event-listeners/onPageComplete.js
```diff
@@ -5,11 +5,19 @@ export default store => pageIndex => {
   if (state.viewer.isAccessibleMode) {
     core.getDocument().loadPageText(pageIndex, text => {
       const textContainer = document.createElement('div');
+      const className = 'accessible-text';
```
ZhijieZhang: Hmm this container has an id `pageText${pageIndex}`, so can we use `getElementById` to select it?

achenPDFTron: yes. thanks for pointing that out. i have made the fix.

PR #440, opened 2019-12-16
Fix issue with only printing the first page in IE and Edge

mparizeau-pdftron:
Only apply Safari workaround on iOS devices as the CSS causes
the printing problem for IE and Edge.

### DIFF
--- a/src/components/PrintHandler/PrintHandler.js
+++ b/src/components/PrintHandler/PrintHandler.js
@@ -15,20 +15,13 @@ const PrintHandler = () => {
     shallowEqual,
   );
 
-  let containerStyle;
+  let containerClass = 'PrintHandler';
   if (isIOS) {
-    // workaround for getting safari to print to the whole page
-    containerStyle = {
-      top: 0,
-      bottom: 0,
-      left: 0,
-      right: 0,
-      position: 'absolute',
-    };
+    containerClass += ' ios-print';
   }
 
   return isDisabled ? null : (
-    <div className="PrintHandler" style={containerStyle}>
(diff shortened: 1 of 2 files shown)


ZhijieZhang:
Hmm from my testing it seems that after canceling the printing on iOS the images will be visible in the viewer. I guess it's because `top`, `left`, `right`, `bottom` always have a value now instead of being `0` only when `@media print`?

Can you reproduce that? If you can reproduce this then I think to solve this issue we can use `window.matchMedia('print')`. Listen to the media change if it's only iOS and when it matches we set the container style.

ZhijieZhang:
One question... Should this be merged into 6.0 instead?

ZhijieZhang:
Actually I googled a bit and found that there might be a css way to do this. Haven't tested how well it worked out https://stackoverflow.com/questions/30102792/css-media-query-to-target-only-ios-devices

src/components/PrintHandler/PrintHandler.js
```diff
@@ -14,8 +15,13 @@ const PrintHandler = () => {
...
   );
 
+  let containerClass = 'PrintHandler';
+  if (isIOS) {
+    containerClass += ' ios-print';
+  }
+
   return isDisabled ? null : (
-    <div className="PrintHandler">
+    <div className={containerClass}>
```
ZhijieZhang: Can we use the classnames package please ;). So it will be like
```javascript
import classNames from 'classnames';

<div className={classNames({
  PrintHandler: true,
  'ios-print': isIOS,
})}>
...
</div>
```

PR #441, opened 2019-12-17
Updated the styles for the thumbnail. Created more margin at the top. Also added a hover background to the container. Refactored the code a bit so that the hardcoded size is only in one place. Before it was in both the components and the style sheet.

ZhijieZhang:


src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -5,7 +5,7 @@ import { List } from 'react-virtualized';
 import Measure from 'react-measure';
 import classNames from 'classnames';
 
-import Thumbnail from 'components/Thumbnail';
+import Thumbnail, { THUMBNAIL_SIZE } from 'components/Thumbnail';
```
awy3: `THUMBNAIL_SIZE ` is null and `getNumberOfColumns` breaks because it comparing with undefined values.  I think in `components/Thumbnail/index.js`, you need to change it to 
```
export { default } from './Thumbnail';
export { THUMBNAIL_SIZE } from './Thumbnail';
```

PR #447, opened 2019-12-20
Feature/measurement overlay updates

bollain:
Requires: This[ branch ](https://github.com/XodoDocs/webviewer/pull/571)from webviewer if not already merged. 

QA Steps: 
- Run webviewer and start using the line measurement annotation tool. 
- In the measurement overlay in the bottom right corner you will now notice the length property is an input. 
- Try changing the value and see how the line redraws. If you enter a large value that causes the line to go off screen, it should automatically resize to stay within bounds when you click anywhere on the document or tab away from the input box.


src/components/MeasurementOverlay/MeasurementOverlay.js
```diff
@@ -221,6 +229,69 @@ class MeasurementOverlay extends React.PureComponent {
     );
   };
 
+  onChangeLineLength = length => {
```
ZhijieZhang: It'd be very nice if we can put these functions into the `LineMeasurementInput` :)

I think what prevents us from doing that is
- In `ComponentDidUpdate` we use `ensureLineIsWithinBounds` and if we move `ensureLineIsWithinBounds` to the `LineMeasurementInput` component we won't have access to it in the parent component(assuming we are not using `useimperativehandle`). This can be solved by passing `isOpen` as a prop down to `LineMeasurementInput` so that we can check this prop in it and call `ensureLineIsWithinBounds` whenever it's necessary
- In some functions we call `forceLineRedraw`, which uses `this.forceUpdate`. If we put these functions into the child component then we don't have access to `this.forceUpdate` anymore. This can be solved by passing `forceUpdate` down to the child component, which I think is a bit weird... Another solution I can think of is to call `annotManager.trigger('annotationChanged', [[lineAnnotation], 'modify', {}])` in the child component whenever we change the length of the line. Since the parent component is listening to `annotationChanged` and calling `this.forceUpdate`, we don't need to pass `forceUpdate` down to the child component anymore. This might actually be important for the realtime-collaboration sample(https://www.pdftron.com/documentation/web/samples/annotation/). This sample is listening to `annotaitonChanged` event to check if any annotations have been modified to sync annotations in two viewer. If we don't trigger `annotationChanged` event when the value of input changes then the line length may be different in two viewers.

@mparizeau-pdftron what do you think? Is it necessary to trigger an `annotationChanged` event here?

mparizeau-pdftron: Yes, I think it would make sense to trigger annotationChanged here since we're changing a property of the annotation and can have it hook into that normal flow.

bollain: Sounds good, I will move that line logic into the `LineMeasurementInput` component.

PR #449, opened 2019-12-23
Bugfix/hot keys arrows

achenPDFTron:


src/helpers/hotkeysManager.js
```diff
@@ -251,6 +253,26 @@ WebViewer(...)
           core.setCurrentPage(currPageNumber + 1);
         }
       },
+      up: e => {
+        // do not call preventDefault else it will prevent scrolling
+        const scrollViewElement = core.getScrollViewElement();
+        const { scrollTop } = scrollViewElement;
+        const reachedTop = scrollTop === 0;
+        const currPageNumber = core.getCurrentPage();
+        if ((e.key === 'ArrowUp' || e.which === 38) && reachedTop && currPageNumber > 1) {
```
ZhijieZhang: I don't think we need to check the `key` and `which` here since the library handles this for us. The library should call this handler only if the `key` is `ArrowUp`

src/helpers/hotkeysManager.js
```diff
@@ -251,6 +253,26 @@ WebViewer(...)
...
+          core.setCurrentPage(currPageNumber - 1);
```
ZhijieZhang: The `this.pageUp` in `DocumentContainer.js` has this line `this.container.current.scrollTop = scrollHeight - clientHeight;`. What it does is that when the page is zoomed in and the scroll bar reaches the top, we will go to the previous page but keep the scroll position to the bottom.

To remain the same functionality this line should be copied over here as well.

src/helpers/hotkeysManager.js
```diff
@@ -33,6 +33,8 @@ const NOOP = () => {};
  * @property {string} Ctrl_P Print
  * @property {string} PageUp Go to the previous page
  * @property {string} PageDown Go to the next page
+ * @property {string} Up Go to the previous page if not scrolling (refers to the ArrowUp key)
```
ZhijieZhang: As described in another comment, the arrow key can also be used when the page is zoomed in(there's a scrollbar). So we need to change the documentation.

How about `Go to the previous page in the single layout mode`? Also I feel like inside the bracket we can just write `ArrowUp` to make it consistent with `Ctrl+Shift+=`.

src/helpers/hotkeysManager.js
achenPDFTron: addressed. thanks for pointing that out. i must have blindly copied and pasted.

src/helpers/hotkeysManager.js
achenPDFTron: addressed

src/helpers/hotkeysManager.js
achenPDFTron: addressed

PR #450, opened 2019-12-28
[Feature] UI support for adding rubber stamp annotations

sisardor:
Initial UI looks like this:

Screen Shot 2019-12-27 at 5 52 06 PM


ZhijieZhang:
Hi Sardor. Let me know when this is ready for review. Thanks!

src/components/ToolStylePopup/ToolStylePopup.js
```diff
@@ -4,7 +4,7 @@ import onClickOutside from 'react-onclickoutside';
 import { connect } from 'react-redux';
 
 import StylePopup from 'components/StylePopup';
-
+import getToolStylePopupPositionBasedOn from 'helpers/getToolStylePopupPositionBasedOn';
```
sisardor: Moved to a module for reusability.

src/event-listeners/onRubberStampAnnotationAdded.js
```diff
@@ -0,0 +1,12 @@
+import core from 'core';
+import defaultTool from 'constants/defaultTool';
+import actions from 'actions';
+
+export default dispatch => rubberStampAnnotation => {
+  core.setToolMode(defaultTool);
+  core.getTool('AnnotationCreateRubberStamp').hidePreview();
+  dispatch(actions.closeElement('cursorOverlay'));
```
ZhijieZhang: This isn't needed now because the logic to show and hide the preview element is in WebViewer. if you merge master into this branch you should see that I deleted this in `onSignatureAnnotationAdded.js`

src/event-listeners/onRubberStampAnnotationAdded.js
```diff
@@ -0,0 +1,12 @@
...
+  setTimeout(() => {
+    core.selectAnnotation(rubberStampAnnotation);
```
ZhijieZhang: I don't think we need to wrap this with a `setTimeout 0`. Can you double check?

src/components/StampOverlay/StampOverlay.scss
```diff
@@ -0,0 +1,48 @@
...
+  width: 232px;
+  border-radius: 4px 0 0 4px;
+  box-shadow: 0 2px 16px 0 #E1E1E3;
+  width: 354px;
+  padding: 16px 16px 16px;
+
+
+  .rubber-stamp {
+    float: left;
+    // background: #f8f8f8;
```
ZhijieZhang: I think we can remove commented code

src/components/StampOverlay/StampOverlay.scss
```diff
@@ -0,0 +1,48 @@
...
+
+    img {
+      width: 100%;
+      height: 100%;
+      cursor: pointer;
+    }
+  }
+  .rubber-stamp:hover {
+    // background: antiquewhite;
+    background: var(--button-hover-color);
```
ZhijieZhang: Does css variable work in IE11?

src/components/StampOverlay/StampOverlay.js
```diff
@@ -0,0 +1,169 @@
...
+      imgs = defaultAnnotations.map(({ imgSrc, annotation }, index) =>
+        <div key={index}
+          className="rubber-stamp"
+          onClick={() => this.setRubberStamp(annotation)}
+        >
+          <img src={imgSrc} />
+        </div>,
+      );
+    }
+    const className = getClassName('Overlay StampOverlay', this.props);
```
ZhijieZhang: Can we use the `classnames` package to handle class names? I'd like to remove the use of `getClassName`. Thanks!

src/components/StampOverlay/StampOverlay.js
```diff
@@ -0,0 +1,169 @@
...
+    closeElement: PropTypes.func.isRequired,
+    closeElements: PropTypes.func.isRequired,
+  };
+
+  constructor(props) {
+    super(props);
+    this.overlay = React.createRef();
+
+    const { toolButtonObjects } = props;
+    const dataElement = toolButtonObjects[TOOL_NAME].dataElement;
```
ZhijieZhang: We can hardcode this but I guess accessing it at runtime is a better practice? :)

src/components/StampOverlay/StampOverlay.js
```diff
@@ -0,0 +1,169 @@
...
+      language: props.i18n.language,
+    };
+    this.stampTool = core.getTool(TOOL_NAME);
+  }
+
+  componentDidUpdate(prevProps) {
+    // if langauge changes while overlay is open we wanted update it
+    var isLanChanged = this.state.language && this.props.i18n.language !== this.state.language;
+
+    if (!prevProps.isOpen && !this.props.isOpen && this.props.activeToolName === TOOL_NAME) {
```
ZhijieZhang: I'm a bit lost here. Can you explain what `!prevProps.isOpen && !this.props.isOpen` does? It seems to me that we want to open this overlay when the previous `activeToolName` isn't `AnnotationCreateRubberStamp` but `this.props.activeToolName` is? If this is the case I think we can just do something like:
```javascript
if (prevProps.activeToolName !== this.props.activeToolName && this.props.activeToolName === TOOL_NAME)
```

src/components/StampOverlay/StampOverlay.js
```diff
@@ -0,0 +1,169 @@
...
+      top: 0,
+      defaultAnnotations: [],
+      dataElement,
+      language: props.i18n.language,
+    };
+    this.stampTool = core.getTool(TOOL_NAME);
+  }
+
+  componentDidUpdate(prevProps) {
+    // if langauge changes while overlay is open we wanted update it
```
ZhijieZhang: Typo. Should be language

src/components/StampOverlay/StampOverlay.scss
sisardor: I changed to 
```
background: #F6F6F6;
background: var(--button-hover-color);
```
so it can fallback to otherone is `var` is not supported
it is used the same way as in Button.scss

ZhijieZhang:
Hi Sardor, I'm still seeing the issue that I pointed out in this comment https://github.com/PDFTron/webviewer-ui/pull/450#pullrequestreview-343005018.

To get rid of the slider I can think a workaround we can do is to change 
https://github.com/PDFTron/webviewer-ui/blob/4837e068c5be156ec58361f63ec7a8d8bf5f906b/src/helpers/toolStylesExist.js#L3-L10

to return false when it's the rubber stamp annotation tool.

I think it's a bit weird to see two overlays(the opacity slider and the stamp overlay) show up at the same time.

To fix the issue where sometimes the stamp overlay doesn't show up a workaround for now is to change the `ToolButton` component to behave a bit different when clicking on a rubber stamp annotation tool.

PR #451, opened 2019-12-30
Misc/use embededed print

achenPDFTron:


achenPDFTron:
What documentation currently looks like:
image

The square bracket gives off a vibe that it takes on an "array" as an argument.
Example:
image


Modified it so that it looks like this:
image

src/apis/useEmbeddedPrint.js
```diff
@@ -3,7 +3,7 @@
  * The printing process will be faster and the quality might be higher when using Chrome's native printing.
  * You may not want to use embedded printing if there are custom annotations in your document.
  * @method WebViewerInstance#useEmbeddedPrint
- * @param {boolean} [use=true] Whether or not to use embedded printing
+ * @param {boolean} use Whether or not to use embedded printing
```
mparizeau-pdftron: @achenPDFTron The square brackets are how we specify that the parameter is optional and "=true" specifies what the default value is so unfortunately I don't think we can or should remove this.

achenPDFTron: so i did some googling around. i found a syntax for declaring default value but i was not able to find another syntax to declare as optional without making it look like an array in the doc. i will leave things as is.

PR #455, opened 2020-01-03
Bug fix/crop tool not disabled

achenPDFTron:


src/helpers/createToolAPI.js
```diff
@@ -33,12 +33,12 @@ export default (enable, store) => (
       hotkeys.on(toolName);
     });
 
-    store.dispatch(actions.enableElements(dataElements, PRIORITY_ONE));
+    store.dispatch(actions.enableElements(dataElements, PRIORITY_THREE));
```
achenPDFTron: i pick to be 3 so that its the highest priority so that it won't ever be enabled.
i feel like if users are calling "disableTools", they definitely want it to be disabled.

ZhijieZhang: Not sure if we should choose `2` or `3`. I feel like only `instance.disableElements`, `instance.enableElements` and the `disabledElements` can have `3` because they are very specific. I will merge it in first. Thanks!

PR #456, opened 2020-01-03
[CHANGE] fixed bug where fields in print modal were not disabled when…

achenPDFTron:
… pages are being processed to print

### DIFF
--- a/src/components/PrintModal/PrintModal.js
+++ b/src/components/PrintModal/PrintModal.js
@@ -52,6 +52,7 @@ class PrintModal extends React.PureComponent {
       isWatermarkModalVisible: false,
       watermarkModalOption: null,
       existingWatermarks: null,
+      isProcessingPagesToPrint: false,
     };
   }
 
@@ -135,7 +136,7 @@ class PrintModal extends React.PureComponent {
       return;
     }
 
-    this.setState({ count: 0 });
+    this.setState({ count: 0, isProcessingPagesToPrint: true });
     this.setPrintQuality();
 
     if (this.state.allowWatermarkModal) {
@@ -149,9 +150,11 @@ class PrintModal extends React.PureComponent {
       .then(pages => {
         this.printPages(pages);
         this.resetPrintQuality();
+        this.setState({ isProcessingPagesToPrint: false });
       })
       .catch(e => {
         console.error(e);
+        this.setState({ isProcessingPagesToPrint: false });
       });
   };
 
@@ -435,14 +438,14 @@ class PrintModal extends React.PureComponent {
   };
 
   closePrintModal = () => {
-    this.setState({ count: -1 });
+    this.setState({ count: -1, isProcessingPagesToPrint: false });
     this.props.closeElement('printModal');
   };
 
   cancelPrint = () => {
     const doc = core.getDocument();
     this.pendingCanvases.forEach(id => doc.cancelLoadCanvas(id));
-    this.setState({ count: -1 });
+    this.setState({ count: -1, isProcessingPagesToPrint: false });
   };
 
   setWatermarkModalVisibility = visible => {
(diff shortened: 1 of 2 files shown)


achenPDFTron:
https://trello.com/c/5qqWF7em/652-print-modal-should-disallow-clicking-on-other-fields-after-print-button-pressed

src/components/PrintModal/PrintModal.js
```diff
@@ -514,35 +521,40 @@ class PrintModal extends React.PureComponent {
...
                 />
               </form>
             </div>
             {this.state.allowWatermarkModal && (
               <button
                 id="applyWatermark"
                 className="apply-watermark"
-                onClick={() => this.setWatermarkModalVisibility(true)}
+                disabled={isProcessingPagesToPrint}
+                onClick={() => !isProcessingPagesToPrint && this.setWatermarkModalVisibility(true)}
```
ZhijieZhang: Can you use a `if` instead of `&&` here? Thanks!

src/components/PrintModal/PrintModal.js
```diff
@@ -497,7 +501,10 @@ class PrintModal extends React.PureComponent {
                 dataElement="printModalCloseButton"
                 title="action.close"
                 img="ic_close_black_24px"
-                onClick={this.closePrintModal}
+                onClick={() => {
+                  this.cancelPrint();
```
ZhijieZhang: If we call `this.cancelPrint` here then we also need to change `line 495` to call `cancelPrint` as well to make the behavior consistent.

src/components/PrintModal/PrintModal.js
```diff
@@ -464,14 +467,15 @@ class PrintModal extends React.PureComponent {
...
     const customPagesLabelElement = (
       <input
         ref={this.customInput}
         type="text"
         placeholder={t('message.customPrintPlaceholder')}
         onFocus={this.onFocus}
+        disabled={this.state.isProcessingPagesToPrint}
       />
     );
     const isPrinting = count >= 0;
```
ZhijieZhang: Is this variable the same as `this.state.isProcessingPagesToPrint`? If it is then we can just use this variable instead of adding a new state.

For example:
```javascript
                <Input
                  ref={this.currentPage}
                  id="current-page"
                  name="pages"
                  type="radio"
                  label={t('option.print.current')}
                  // disabled={isProcessingPagesToPrint}
                  disabled={isPrinting}
                />
```

src/components/PrintModal/PrintModal.js
achenPDFTron: done

src/components/PrintModal/PrintModal.js
achenPDFTron: done

src/components/PrintModal/PrintModal.js
achenPDFTron: done. i'm guessing its because we don'ot want to force a whole rerender when i change the state?

ZhijieZhang: Not really. I find that deriving a state(`isPrinting`) from an existing state(`this.state.count`) is usually better than syncing two states(`this.state.count` and `this.state.isProcessingPagesToPrint`).

What I'm suggesting is that we can just get rid of `isProcessingPagesToPrint` and use the derived `isPrinting` in the render function, as the above code snippet showed. It seems to me that `isPrinting` is the same as `isProcessingPagesToPrint` so we can just use `isPrinting`.

achenPDFTron: done. funny how i didn't see `isPrinting`.

I have a question if we go this route.
image
Should we set `this.state.count` to be -1 on error?

I don't know how to force an error. But I'm guessing it will remain disabled even on error which could be a problem in terms of user experience.

ZhijieZhang: Uhh I missed this comment. Yeah we should set it to -1 in the catch block. I will do it.

PR #460, opened 2020-01-07
Removed index.build.html and changed prod script to use index.core.html

achenPDFTron:


achenPDFTron:
Difference between the 2 files:
image

Removed and modified to fix this bug:
https://pdftron.freshdesk.com/a/tickets/9986

PR #461, opened 2020-01-07
Feat/custom measurement overlay

bollain:
- Adds new interface to create custom measurement overlays. 
- To start we have added this functionality to the Ellipsis annotation.
- To get a custom overlay a customer would need to invoke the following method:

```
    instance.setCustomMeasurementOverlayInfo([
    {
      title: "Radius Measurement",
      label: "Radius",
      validate: annotation => annotation instanceof instance.Annotations.EllipseAnnotation,
      value: annotation => annotation.Width / 2,
      onChange: (e, annotation) => {
        // Do something with the annot like resize
        console.log(annotation)
        console.log('event ', e.target.value)
      }
    }
  ])
```

It takes an array with configuration objects for the custom overlay. At this time we are only testing it with Ellipsis; the interface is still very much fluid and it's open to discussion. 
Once this parameter is passed an overlay will pop up when the ellipse annotation is being drawn, and when it is selected.


src/components/MeasurementOverlay/CustomMeasurementOverlay.js
```diff
@@ -0,0 +1,75 @@
...
+function CustomMeasurementOverlay(props) {
+  const renderAppropriateOverlay = type => {
+    const annotationType = mapAnnotationToKey(props.annotation);
+    switch (annotationType) {
+      case 'ellipse':
+        return <CustomEllipseMeasurementOverlay {...props}/>;
+      default:
+        console.error(`Custom overlay for annotation type: ${type} not supported`);
+    }
+  };
```
bollain: I found that if we end up implementing overlays for other annotations they will have some variation... and it's hard to come up with a base component to cover all cases. Therefore I think this "factory" approach is best, but maybe there is a better way?

bollain:
Added docs, let me know if this is clear and concise:
image

PR #465, opened 2020-01-09
[CHANGE] Allow instance.updateElement to change the colors in the color palette

ZhijieZhang:
QA:
- In the console, do `readerControl.updateElement('colorPalette', ['#FFFFFF'])`
- Now the palette should only contain the white color

This is a request from a customer who wishes to customize the colors that are available in the color palette.  There're some noises in this PR since I formatted the component using Prettier, and there're some refactoring as well. I have to do some refactoring because the previous implementation has some hardcoded values that don't work well anymore when users are able to customize the colors.

### DIFF
--- a/src/apis/updateElement.js
+++ b/src/apis/updateElement.js
@@ -1,10 +1,10 @@
 import actions from 'actions';
 
 /**
- * Update an element in the viewer. Currently this API can only update elements that have a Button class name.
+ * Update an element in the viewer.
  * @method WebViewerInstance#updateElement
- * @param {string} dataElement the data element of the element that will be updated.
- * @param {object} props An object that is used to override an existing item's properties.
+ * @param {string} dataElement the data element of the element that will be updated. Valid values are 'colorPalette', and HTML elements that have 'Button' in the class name.
+ * @param {*} props An object or an array that is used to override an existing item's properties.
  * @example
 WebViewer(...)
   .then(function(instance) {
@@ -12,6 +12,8 @@ WebViewer(...)
       img: 'path/to/image',
       title: 'new_tooltip',
     })
+
+    instance.updateElement('colorPalette', ['#FFFFFF', 'transparency', '#000000'])
   });
  */
 export default store => (dataElement, overrides) => {
--- a/src/apis/updateElement.js
+++ b/src/apis/updateElement.js
@@ -15,5 +15,44 @@ WebViewer(...)
   });
  */
 export default store => (dataElement, overrides) => {
-  store.dispatch(actions.setCustomElementOverrides(dataElement, overrides));
+  switch (dataElement) {
+    case 'colorPalette':
+      overrides = validateColorPaletteOverrides(overrides);
+      break;
+    default:
+      overrides = validateButtonOverrides(overrides);
+  }
+
+  if (overrides) {
+    store.dispatch(actions.setCustomElementOverrides(dataElement, overrides));
+  }
+};
+
+const validateColorPaletteOverrides = overrides => {
+  if (!Array.isArray(overrides)) {
+    return console.warn(
+      'The second argument needs to be an array of strings to update the color palette',
+    );
+  }
+
+  // examples of valid colors are: '#f0f0f0', '#FFFFFF'
+  const isValidColor = color =>
+    color === 'transparency' ||
+    (color.startsWith('#') && color.split('#')[1].length === 6);
+
+  if (!overrides.every(isValidColor)) {
+    return console.warn('The color must be \'transparency\' or a hex color string. For example #F0F0F0');
+  }
+
+  return overrides;
+};
+
+const validateButtonOverrides = overrides => {
+  if (overrides !== null && typeof overrides !== 'object') {
+    return console.warn(
+      'The second argument needs to be an object to update a button',
+    );
+  }
+
+  return overrides;
 };
--- a/src/apis/updateElement.js
+++ b/src/apis/updateElement.js
@@ -15,9 +15,5 @@ WebViewer(...)
   });
  */
 export default store => (dataElement, overrides) => {
-  if (overrides !== null && typeof overrides !== 'object') {
-    return console.warn('The second argument to instance.updateElement needs to be an object.');
-  }
-
   store.dispatch(actions.setCustomElementOverrides(dataElement, overrides));
 };
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { useSelector } from 'react-redux';
+import { useSelector, shallowEqual } from 'react-redux';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
 
@@ -11,7 +11,6 @@ import selectors from 'selectors';
 import './Button.scss';
 
 const propTypes = {
-  disable: PropTypes.bool,
   isActive: PropTypes.bool,
   mediaQueryClassName: PropTypes.string,
   img: PropTypes.string,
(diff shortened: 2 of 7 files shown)


src/apis/updateElement.js
```diff
@@ -15,9 +15,5 @@ WebViewer(...)
   });
  */
 export default store => (dataElement, overrides) => {
-  if (overrides !== null && typeof overrides !== 'object') {
```
achenPDFTron: will this API be the only one where its second param is of an array?

src/components/Button/Button.js
```diff
@@ -11,7 +11,6 @@ import selectors from 'selectors';
 import './Button.scss';
 
 const propTypes = {
-  disable: PropTypes.bool,
```
achenPDFTron: this might cause a bug.
I recently added this so that we can disable the print button when "printing is in progress".

I think we should keep it?

Reasoning:
Looking at the name, it looks like be a general reusuable dumb presentational button component.
It looks like a bit strange that we need to read from the state to know if its disabled or not?

I don't quite know the whole "knowledge" of this file. so let me know if i'm missing something.

src/components/ColorPalette/ColorPalette.js
```diff
@@ -1,92 +1,164 @@
...
+    '#373737',
+    '#000000',
+  ];
 
   setColor = e => {
     const { property, onStyleChange } = this.props;
     const bg = e.target.style.backgroundColor; // rgb(r, g, b);
-    const rgba = bg ? bg.slice(bg.indexOf('(') + 1, -1).split(',') : [0, 0, 0, 0];
-    const color = new window.Annotations.Color(rgba[0], rgba[1], rgba[2], rgba[3]);
+    const rgba = bg
```
achenPDFTron: looks like reformatting?

src/components/ColorPalette/ColorPalette.js
```diff
@@ -1,92 +1,164 @@
...
-  renderTransparencyCell = (bg, key) => {
+  renderTransparencyCell = bg => {
     const { property } = this.props;
-    const shouldRenderDummyCell = property === 'TextColor' || property === 'StrokeColor';
+    const shouldRenderDummyCell =
+      property === 'TextColor' || property === 'StrokeColor';
 
     if (shouldRenderDummyCell) {
-      return <div className="dummy-cell" key={key}></div>;
+      return <div className="dummy-cell" />;
```
achenPDFTron: i'm still getting the hang of react.
what is the difference between the two? I'm guessing one just doesn't need the key? but both renders it as an HTML element?

achenPDFTron:
@ZhijieZhang as discussed and found in person.

Steps to reproduce bug
1. call the api: readerControl.updateElement('colorPalette', ['#FFFFFF'])
2.  use the UI to create a new free text annotation
3. type something
4. click on the free text annotation to edit it
5. choose "fill color"

Error occurs:
image

ZhijieZhang:
> Works as intended! Would there be a point to potentially validate the pallete being passed in? Or would that . be too complicated/unnecessary?

I added a bit validation to the palette input. https://github.com/PDFTron/webviewer-ui/blob/d44daa6772ef0c3aba22dce98956b2362398e4d1/src/apis/updateElement.js#L34-L49
Let me know if you think this is ok. We can potentially allow the hex color string to be a shorthand(`#fff` instead of `#ffffff`). I think what we have is ok for now. If this is requested then for sure we will add it.

src/apis/updateElement.js
ZhijieZhang: Yes it is. I updated the code to validate the input.

src/components/Button/Button.js
ZhijieZhang: I checked the commit history and it shows that I added this prop a few months ago. I don't think this prop is used anymore in this component so I just removed it.

I think this is the one that is added by you? https://github.com/PDFTron/webviewer-ui/blob/master/src/components/Input/Input.js#L17

src/components/ColorPalette/ColorPalette.js
ZhijieZhang: Yes!

src/components/ColorPalette/ColorPalette.js
ZhijieZhang: These two doesn't have any difference. The new one is a shortcut. We can use the shortcut if the element we are rendering doesn't have a child.

I didn't use the `key` for this `div` because I refactored the code to use key here https://github.com/PDFTron/webviewer-ui/blob/master/src/components/Input/Input.js#L17.

ZhijieZhang:
> @ZhijieZhang as discussed and found in person.
> 
> Steps to reproduce bug
> 
> 1. call the api: readerControl.updateElement('colorPalette', ['#FFFFFF'])
> 2. use the UI to create a new free text annotation
> 3. type something
> 4. click on the free text annotation to edit it
> 5. choose "fill color"
> 
> Error occurs:
> image

Thanks for catching this! I fixed this in the new commits. https://github.com/PDFTron/webviewer-ui/blob/d44daa6772ef0c3aba22dce98956b2362398e4d1/src/components/ColorPalette/ColorPalette.js#L128-L132

src/apis/updateElement.js
achenPDFTron: doesn't it feel weird that this accepts 2 kinds of inputs?
An object with keys or an array?

Wouldn't it be a better idea to allow it to accept a parameter of one type?

src/components/Button/Button.js
achenPDFTron: ah yes. thanks for the correction.

achenPDFTron:
@ZhijieZhang just 1 more comment on my end

src/apis/updateElement.js
ZhijieZhang: @achenPDFTron  I'm not sure if it's weird to users. I think function overloading is common to see for APIs. I guess the API signature should be whatever makes more sense to the user. If the second argument has to be an object, how would they use the API to update the palettes? Maybe something like
```javascript
instance.updateElement('colorPalette', {
  colors: ['#FFFFFF', '#...']
})
```

This can definitely work but I just feel like passing an array of colors is more intuitive...

achenPDFTron: no i agree with you. passing an array of strings sounds more intuitive and easier in this case.

I think we should also consider how easy is it for us to maintain the API as well; not just whatever makes sense to the users.

My concern is how the second parameter is either an array or an object and how this could make it harder to debug and understand.

Ideally we should use an object or create a new function.

It looks like you have comments and documentations for now which could work.

I think in the far future, if we have more customize-able properties for the color palette, I think we should consider using an object.

ZhijieZhang: I agree what you said. Not sure if you've checked the latest code yet but I added some validations and break the main function body of `updateElement` into some small functions.

```javascript
switch (dataElement) {
  case 'colorPalette':
    overrides = validateColorPaletteOverrides(overrides);
    break;
  default:
    overrides = validateButtonOverrides(overrides);
}

if (overrides) {
  store.dispatch(actions.setCustomElementOverrides(dataElement, overrides));
}
```

I hope this can make it a bit clear about how this API is used? So to add more elements we just add more cases. For now I don't think it will be difficult to maintain because essentially this API just dispatch an action to the store and components just connect to that state to see if there're any custom overrides.

I agree that this API may get harder to harder to document once more elements can be customized with different way. So to make documentation easier and more clear I think we can split them into different APIs if necessary.

At least for me It's pretty hard to predict how things will change so I think if you are not strongly against the way the API(and the code structure) is written, we can let this PR go through.

achenPDFTron: i think I am fine with it for now. i think if we have one more different dataElement type, we might want to rethink think.

ZhijieZhang:
I'm going to merge this one in master. Let me know if there're any questions.

PR #473, opened 2020-01-17
[Feature ]Added URL/Page link tool

andreysaf:
- You are now able to link text selection or any annotation to a URL or a page.
- If the selected text contains URL, it will be automatically prepopulated.
- Autofocus on URL/Page Number select.
- If the annotation or highlight gets deleted, the associated link will be deleted as well.

Related PR in WebViewer https://github.com/XodoDocs/webviewer/pull/690


src/components/TextPopup/TextPopup.js
```diff
@@ -114,6 +114,14 @@ const TextPopup = () => {
           }
           dataElement="textStrikeoutToolButton"
         />
+        <ActionButton
```
ZhijieZhang: Do you want to make this as a separate component, say `LinkButton`, and import this button in `AnnotationPopup` and `TextPopup`? This way we can avoid some duplication. I'm ok with what we have now.

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+      //  prepopulate page numbers for the dropdown
+      const numbers = [];
+      for (let i = 1; i <= totalPages; i++) {
+        numbers.push(
+          <option key={i} value={i}>
+            {i}
+          </option>,
+        );
+      }
+      setDropdownNumbers(numbers);
```
ZhijieZhang: Instead of storing `<option>`s on a state can we just have them in the returned JSX?
```jsx
                <select
                  className="pageNumberSelect"
                  ref={pageNumberInput}
                  value={pageNumber}
                  onChange={e => setPageNumber(e.target.value)}
                >
                  {/* create <option>s here based on the value of totalPages */}
                </select>
```

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+    highlight.Author = core.getCurrentUser();
+    highlight.setContents(text);
+
+    let linkAnnotIdArray = [];
+    linkAnnotArray.forEach(link => {
+      linkAnnotIdArray.push(link.Id);
+    });
+    highlight.associateLink(linkAnnotIdArray);
+
+    await core.addAnnotations([highlight]);
```
ZhijieZhang: Does `addAnnotations` return a promise? I'm not sure if we need to await this function and `drawAnnotationsFromList`

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+                <select
+                  className="pageNumberSelect"
+                  ref={pageNumberInput}
+                  value={pageNumber}
+                  onChange={e => setPageNumber(e.target.value)}
+                >
+                  {dropdownNumbers}
+                </select>
+              </div>
+              <div className="buttons">
```
ZhijieZhang: Is it necessary to have this `div`? Also I don't think we are using this class name in `LinkModal.scss` so maybe we can just remove it?

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+                  dataElement="linkSubmitButton"
+                  label={t('action.link')}
+                  onClick={addURLLink}
+                />
+              </div>
+            </form>
+          </TabPanel>
+          <TabPanel dataElement="PageNumberPanel">
+            <form onSubmit={addPageLink}>
+              <div className="enter">
```
ZhijieZhang: Do we use this class name?

i18n/translation-en.json
```diff
@@ -252,6 +254,13 @@
   "tool": {
     "pan": "Pan",
     "select": "Select",
-    "Marquee": "Marquee Zoom"
+    "Marquee": "Marquee Zoom",
+    "Link": "Link URL or Page"
+  },
+  "link": {
+    "URL": "URL",
+    "Page": "Page",
```
ZhijieZhang: to make things consistent can we make them lowercased just as other keys?

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+    const quads = core.getSelectedTextQuads();
+    const selectedAnnotations = core.getSelectedAnnotations();
+
+    const currPageQuads = quads[currentPage - 1];
+
+    if (currPageQuads) {
+      const selectedText = core.getSelectedText();
+      currPageQuads.forEach(quad => {
+        const link = new Annotations.Link();
+        link.PageNumber = currentPage;
```
ZhijieZhang: It'd be good if we can remove the duplicate code here. we can potentially have a function that takes the `X`, `Y`, `Width` and `Height` and use that function here and inside the `if` block below.

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+
+  const createHighlightAnnot = async(linkAnnotArray, quads, text) => {
+    const linkAnnot = linkAnnotArray[0];
+    const highlight = new Annotations.TextHighlightAnnotation();
+    highlight.PageNumber = linkAnnot.PageNumber;
+    highlight.X = linkAnnot.X;
+    highlight.Y = linkAnnot.Y;
+    highlight.Width = linkAnnot.Width;
+    highlight.Height = linkAnnot.Height;
+    highlight.StrokeColor = new Annotations.Color(0, 0, 0, 0);
```
ZhijieZhang: Can you remind me why we want to programmatically create a highlight annotation here? I remember we discussed about this a bit but I don't remember... Sorry

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+    await core.drawAnnotationsFromList([highlight]);
+  };
+
+  const addURLLink = () => {
+    const links = createLink();
+
+    const action = new window.Actions.URI({ uri: url });
+    links.forEach(async link => {
+      link.addAction('U', action);
+      await core.addAnnotations([link]);
```
ZhijieZhang: Same here

src/components/LinkModal/LinkModal.js
```diff
@@ -0,0 +1,253 @@
...
+    const links = createLink();
+
+    const Dest = window.Actions.GoTo.Dest;
+    const options = { dest: new Dest({ page: pageNumber }) };
+    const action = new window.Actions.GoTo(options);
+
+    links.forEach(async link => {
+      link.addAction('U', action);
+      await core.addAnnotations([link]);
+      await core.drawAnnotationsFromList([link]);
```
ZhijieZhang: Same here

src/components/TextPopup/TextPopup.js
andreysaf: I think ActionButton is fine, since we don't anticipate it changing much and not much different from other action button. Otherwise we would duplicate code and create action button, redaction button, link button?

src/components/LinkModal/LinkModal.js
andreysaf: Fixed that

src/components/LinkModal/LinkModal.js
andreysaf: await was necessary on that one, API doc didn't say that but after I removed it, my link annots stopped rendering.

src/components/LinkModal/LinkModal.js
andreysaf: Resolved

src/components/LinkModal/LinkModal.js
andreysaf: Resolved

i18n/translation-en.json
andreysaf: Fixed

src/components/LinkModal/LinkModal.js
andreysaf: resolved

src/components/LinkModal/LinkModal.js
andreysaf: we wanted to have consistent behaviour for deleting annots, for example create a shape link it, delete shape and link goes away. Selecting text didnt have associated shape or highlight. We create invisible highlight.

src/components/LinkModal/LinkModal.js
andreysaf: await needed :/

src/components/LinkModal/LinkModal.js
andreysaf: await needed :/

src/components/TextPopup/TextPopup.js
ZhijieZhang: Hmm I meant to say that we can create a LinkButton component that renders an ActionButton and use that LinkButton in the two Popup components. So if we ever change the props to this LinkButton we just need to change 1 place. But just like what you said we don't expect this button to be changed much so what we have is totally Okay

src/components/LinkModal/LinkModal.js
ZhijieZhang: I can reproduce the issue you were having. I tried to remove the `await`s and can only see the link annotations appear after I zoom in. I still didn't figure out why but I'm a bit concerned about using `await` to make the code work since apparently  `annotManager.addAnnotations` is synchronous and doesn't return a promise.

After checking the code it seems that link annotations is a HTML annotation, and to draw it we need to call `annotManager.drawAnnotations` with the third argument being true. The third argument is `majorRedraw` and when it's true it will redraw HTML annotations such as link annotations. 

However even if I ask annotManager to major redraw all the annotations in the page I still didn't see the link annotations appear... I found that If I don't draw the highlight annotations before drawing the link annotations everything will be good. This is the code I have

```javascript
const createHighlightAnnot = (linkAnnotArray, quads, text) => {
  const linkAnnot = linkAnnotArray[0];
  const highlight = new Annotations.TextHighlightAnnotation();
  highlight.PageNumber = linkAnnot.PageNumber;
  highlight.X = linkAnnot.X;
  highlight.Y = linkAnnot.Y;
  highlight.Width = linkAnnot.Width;
  highlight.Height = linkAnnot.Height;
  highlight.StrokeColor = new Annotations.Color(0, 0, 0, 0);
  highlight.Quads = quads;
  highlight.Author = core.getCurrentUser();
  highlight.setContents(text);

  const linkAnnotIdArray = linkAnnotArray.map(link => link.Id);
  highlight.associateLink(linkAnnotIdArray);

  core.addAnnotations([highlight]);
  // core.drawAnnotationsFromList([highlight]);
};

const addURLLink = () => {
  const links = createLink();

  const action = new window.Actions.URI({ uri: url });
  links.forEach(link => {
    link.addAction('U', action);
    core.addAnnotations([link]);
    core.drawAnnotations(link.PageNumber, null, true);
  });

  closeModal();
};
```

Not sure if this is better though...

ZhijieZhang:
Hmm found another issue. The screenshot should explain it:

same-page

I think we should see the underline on the second page as well? It seems that we intentionally create link annotations for the current page... Let me know if this is intended. Sorry for not catching this in the first round of reviewing!

src/components/LinkModal/LinkModal.js
andreysaf: Resolved, thx Z.

src/components/LinkModal/LinkModal.js
andreysaf: fixed the awaits

src/components/LinkModal/LinkModal.js
andreysaf: fixed the awaits

Issue #359, opened 2019-09-17
Cannot have image and label in Button.js


henry-chris:
Looks like in 5.2 you changed where a button could not have an image and a label at the same time.  I have a button in the header which displays # of concurrent users in a document.  It's not the best, but it works.
![image](https://user-images.githubusercontent.com/4165721/65067457-2c91b800-d94c-11e9-9d8f-b1e00a7731cc.png)

After updating to 5.2 I cannot do that using your built in system.  Possible to get that changed back?  I'm currently just changing the file to get it to work for now.

mparizeau-pdftron:
It looks like the relevant change was here https://github.com/PDFTron/webviewer-ui/commit/7b888c1f67f204f84d7b5cba02c603a615a104f4#diff-d1110304895e39c6e9fd2be689023291

I believe the logic was being cleaned up as part of refactoring and the original intent wasn't to have both at the same time. However it seems reasonable that you might want to have a button that has an icon and a label at the same time.

We'll discuss internally about allowing this behavior explicitly and making sure it's documented correctly.

ZhijieZhang:
We decided to bring it back as we noticed that in our new UI design we actually have cases where a label and an image are both used. The change will be available in 6.1, which we aim to release at the end of this month. 

Let me know if you'd like me to back merge this change to the version you are using! :)

henry-chris:
Nope, it's fine.  I just put it back in manually.

ZhijieZhang:
Thanks. I will close this issue for now as it's been addressed in 6.1. Feel free to follow up if you have any other questions. Thanks!

PR #475, opened 2020-01-18
[CHANGE] Add a 'disable' option to instance.updateElement. When this option is true, HTML elements that have Button in the class name won't be clickable

ZhijieZhang:
This PR also includes the change to allow the Button component to render both image and label at the same time. This is the behavior that we have in 5.1. Reverting to this behavior because there'll be a use case for this in the new UI design, and resolves https://github.com/PDFTron/webviewer-ui/issues/359


src/components/Button/Button.js
```diff
@@ -23,48 +23,47 @@ const propTypes = {
...
   const {
+    disable,
     isActive,
     mediaQueryClassName,
     img,
     label,
     color,
     dataElement,
-    onClick = () => {},
+    onClick = NOOP,
```
achenPDFTron: why not let it have the value of undefined?

src/components/Button/Button.scss
```diff
@@ -26,7 +23,8 @@
     }
   }
 
-  &:hover {
+  &:hover:not(.disable) {
```
achenPDFTron: maybe we might want to check with the design team. when we hover our mouse cursor over the disabled button, should it show a different mouse cursor?

src/components/Button/Button.js
```diff
@@ -23,48 +23,47 @@ const propTypes = {
...
+    <div
+      className={classNames({
+        Button: true,
+        active: isActive,
+        disable,
+        [mediaQueryClassName]: mediaQueryClassName,
+        [className]: className,
+      })}
+      data-element={dataElement}
+      onClick={disable ? NOOP : onClick}
```
achenPDFTron: assuming we go with the way of making it have an initial value of undefined.
can we do:
`!disable && onClick` ?

src/components/Button/Button.js
```diff
@@ -23,48 +23,47 @@ const propTypes = {
...
+
 const Button = props => {
   const [isElementDisabled, customOverrides] = useSelector(state => [
     selectors.isElementDisabled(state, props.dataElement),
     selectors.getCustomElementOverrides(state, props.dataElement),
   ]);
 
-  props = { ...props, ...customOverrides };
   const {
+    disable,
```
achenPDFTron: back when i worked with Angular 3rd party libraries. some values of "undefined" = true. what is the plan for this?

src/components/Button/Button.js
```diff
@@ -23,48 +23,47 @@ const propTypes = {
   onClick: PropTypes.func.isRequired,
 };
 
+const NOOP = () => {};
+
 const Button = props => {
   const [isElementDisabled, customOverrides] = useSelector(state => [
```
achenPDFTron: what is different between the `disabled` props and `isElementDisabled`?

I'm guessing `isElementDisabled` takes higher priority over "disabled" ?

src/components/Button/Button.js
achenPDFTron: don't forget to update the documentation / guides / dev site in case you haven't done so

achenPDFTron:
@ZhijieZhang  we might want to discuss with the design team.
image

I have disabled the button.
Might be good "user/ux" experience to add CSS to make it "look" disabled.

Right, the disabled/non-disabled state looks the same.

This was something my previous team did back when we were designing our custom Angular component library.

achenPDFTron:
@ZhijieZhang comments provided.

src/components/Button/Button.js
ZhijieZhang: `isElementDisabled` will be true if this element is disabled by `instance.disableElements` so this will mean that we don't render this 'button' at all.

When `disable` is true we just make it not clickable.

I can see that the name is confusing. Do you want to use a variable name other than `isElementDisabled` just in this component to make the difference a bit clear? Let me know what names you want to use.

Unfortunately I don't think we can rename `instance.disableElements` to something else for now because this API(or the `disabledElements` constructor option) is used a lot...

src/components/Button/Button.js
ZhijieZhang: I'm not sure what you mean by `some values of "undefined" = true`. Do you have any examples that I can look at?

src/components/Button/Button.js
ZhijieZhang: Leave it as undefined can also work. If we don't have a default value for this then the code will be something like this I think:
```javascript
const { onClick } = { ...props, ...customOverrides };

onClick={(e) => {
  if (!disabled && onClick) {
    onClick(e);
  }
}}
```
If we have a default value then
```javascript
const { onClick = NOOP } = { ...props, ...customOverrides };

onClick={disable ? NOOP : onClick}
```

I think it's more of a personal perference to think which one is better(or cleaner). I can't I like which one more so if the current way works I'm just usually too lazy to change it :)

src/components/Button/Button.js
ZhijieZhang: I answered this in another comment

ZhijieZhang:
> @ZhijieZhang we might want to discuss with the design team.
> image
> 
> I have disabled the button.
> Might be good "user/ux" experience to add CSS to make it "look" disabled.
> 
> Right, the disabled/non-disabled state looks the same.
> 
> This was something my previous team did back when we were designing our custom Angular component library.

Hmm if the button is disabled then the cursor isn't a point so I guess that can sort of indicate the button is disabled? Maybe I can play around with the color as well, to make it look more gray-ish when it's disabled.

ZhijieZhang:
Untitled

I made them look light gray when disabled. Let me know if you think this is ok!

src/components/Button/Button.js
achenPDFTron: looking around it seems I can't seem to remember the examples I found in the past.

From thinking around, undefined usually means false.

src/components/Button/Button.js
achenPDFTron: Seeing how people fork this repo and may potentially touch this code.
Lets use a variable name to make things more clear.

Lets call it ` hideElement` instead of `isElementDisabled` ?

achenPDFTron:
> Untitled
> 
> I made them look light gray when disabled. Let me know if you think this is ok!

@ZhijieZhang Looks fine.

Minor issue:
Color may conflict with:
image

src/components/Button/Button.js
achenPDFTron: hmmmm i'm wondering. if it's disabled, and lets say we have a hotkey associated with it, will it still be triggered by hotkey?

ZhijieZhang:
Alright I implemented what we've discussed.

- When the button is disabled the cursor is different
- When the button is disabled we don't show the tooltip on hover
- When the button is disabled the hotkey is turned off

Let me know if you think it's good enough to merge it in.

src/components/Button/Button.js
achenPDFTron: discussed in person. we should disable hotkey

src/components/Button/Button.scss
achenPDFTron: i see that the cursor becomes disabled

src/components/Button/Button.js
achenPDFTron: i see that its renamed to `removeElement`

achenPDFTron:
@ZhijieZhang looks good. i have done some testing as well. please make sure you update the docs / guides about this new feature if necessary

PR #479, opened 2020-01-21
[FEATURE] Add new style option tool for rectangles to choose style

bollain:
- Add new tool style popup to change the rectangle tool between regular and cloudy style. Make sure you have [this branch ](https://github.com/XodoDocs/webviewer/pull/709)checked out and that you follow the instructions on there.

### DIFF
--- a/src/components/StyleOption/StyleOption.js
+++ b/src/components/StyleOption/StyleOption.js
@@ -1,0 +1,43 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+import './StyleOption.scss';
+
+function StyleOption(props) {
+  const [style, setStyle] = useState(props.initialStyle);
+  const styleOptions = ['regular', 'cloudy'];
+
+  const onChange = value => {
+    setStyle(value);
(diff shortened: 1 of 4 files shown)


src/components/StyleOption/StyleOption.js
```diff
@@ -0,0 +1,44 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+import './StyleOption.scss';
+
+function StyleOption(props) {
+  const initialStyle = props.initialStyle ? props.initialStyle : 'regular';
+  const [style, setStyle] = useState(initialStyle);
```
ZhijieZhang: I haven't checked out the code yet but I'm wondering if we need to have a state in this case. When the component calls `onStyleChange` this component should be rerendered with the props that have the latest `Style`, and then we can just use that `Style` to be the value of `<select />`.

bollain: This is a good catch... I think Im just over-eager to use hooks.

bollain: Removed state as was redundant.

PR #488, opened 2020-01-31
Changed error modal for upcoming error info changes

S0ulDrag0n:
The error message was changed from a string to an object in a PR in the core. The error modal will need to know how to get the message and display it by default.

### DIFF
--- a/src/components/ErrorModal/ErrorModal.js
+++ b/src/components/ErrorModal/ErrorModal.js
@@ -35,7 +35,7 @@ const ErrorModal = () => {
 
   useEffect(() => {
     const onError = error => {
-      error = error.detail || error.message;
+      error = error.detail.message || error.detail || error.message;
       let errorMessage;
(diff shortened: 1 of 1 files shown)


src/components/ErrorModal/ErrorModal.js
```diff
@@ -35,7 +35,7 @@ const ErrorModal = () => {
 
   useEffect(() => {
     const onError = error => {
-      error = error.detail || error.message;
+      error = error.detail.message || error.detail || error.message;
```
ZhijieZhang: Will `error.detail` be `undefined` or `null`? I guess it will be an empty object at least since it's from a DOM CustomEvent?

S0ulDrag0n: Hm, it is null by default but I believe we are sending a message most of the time. I will add a better check just in case.

Issue #401, opened 2019-11-06
Disable switching pages


yty-agnitio:
How can I disable switching pages for the users (since I will be switching the pages for them remotely using WebSockets)?

ZhijieZhang:
Hi,

You can comment this line out https://github.com/PDFTron/webviewer-ui/blob/master/src/components/DocumentContainer/DocumentContainer.js#L101 to disable scrolling pages in the single display mode. And then you can use setLayoutMode to set the layout mode to be single on every document loaded.

```javascript
instance.docViewer.on('documentLoaded', () => {
  instance.setLayoutMode(instance.LayoutMode.Single);
});
```

ZhijieZhang:
@yty-agnitio Just wanna follow up to see how this works for you. Let me know if you have any other questions. Happy to help!

yty-agnitio:
@ZhijieZhang, It worked. Thank you!

ZhijieZhang:
Thanks for confirming. I will close this issue now. Feel free to reopen it if you have any further questions. Thanks!

yty-agnitio:
Hi again.
I just realized that I also need to prevent some users from swiping pages as well. 
Actually I need to prevent all possible ways of switching pages but only for some users (programmatically on condition). Is there any api for that, or please advise how I can achieve that.

ZhijieZhang:
Hello!

Sorry for the delay in the response. Currently there's no API to disable swiping to go to different pages so you need to modify the UI source code.

The code you'd want to comment out is below
https://github.com/PDFTron/webviewer-ui/blob/75d805a2f8650401f34f85edbb62dcc29b8534fa/src/helpers/TouchEventManager.js#L210-L214

I think it makes sense that we can provide an API to disable all the ways to navigate to different pages. This includes `arrowUp`, `arrowDown`, `wheel`, `swipe`. Not sure if we should remove the button that is used for changing the layout mode to continuous scrolling though. Are you using single page layout mode in your app, and never use continuous scrolling?

Adding this API shouldn't be a tricky change but currently we are working on testing for 6.1 release so I don't have a timeline for this yet. Let me know if you have any concerns or questions. Thanks!

ZhijieZhang:
Hello there. We've added an option to disable all the page navigations in WebViewer.

```javascript
instance.disableFeatures([instance.Feature.PageNavigation])
```

[...]

PR #496, opened 2020-02-04
[CHANGE] Added instance.Feature.PageNavigation that can be used to enable/disable all the ways that navigate the viewer to a different page

ZhijieZhang:
This is an attempt to address #401 and https://trello.com/c/pMfbUglx/645-new-api-to-override-mouse-wheel. This PR doesn't explicitly expose an API to disable mouse wheel, but I think in many cases if the customer wants to disable wheel then he probably wants to disable other ways that can change the page number.

To verify:
In console, do instance.disableFeatures([instance.Feature.PageNavigation]), and verify that the page number can't be changed by using wheel, the swiping gesture, arrow up and arrow down.


src/helpers/createFeatureAPI.js
```diff
@@ -1,7 +1,8 @@
 import core from 'core';
 import hotkeys from 'src/apis/hotkeys';
 import localStorageManager from 'helpers/localStorageManager';
-import TouchEventManager from 'helpers/TouchEventManager';
+import touchEventManager from 'helpers/TouchEventManager';
```
achenPDFTron: seeing `hotkeysManager` is camel cased, should we do the same with `TouchEventManager` to make it consistent?

achenPDFTron:
@ZhijieZhang when im in non-mobile mode, pressing page - up and page -down allows me to transition between page?

src/helpers/createFeatureAPI.js
```diff
@@ -129,6 +130,27 @@ export default (enable, store) => features => {
         store.dispatch(actions.setThumbnailMultiselect(enable));
       },
     },
+    [Feature.PageNavigation]: {
+      fn: () => {
+        const { up: arrowUpHandler, down: arrowDownHandler } = hotkeysManager.keyHandlerMap;
```
achenPDFTron: i think we need to add in page up and page down as well? as i can still navigate the page?

achenPDFTron:
@ZhijieZhang thats all my comments for now

src/constants/feature.js
```diff
@@ -14,6 +14,7 @@
  * @property {string} Copy Ability to copy text or annotations use Ctrl/Cmd + C hotkeys or the copy button.
  * @property {string} ThumbnailMerging Ability to drag and drop a file into the thumbnail panel to merge
  * @property {string} ThumbnailReordering Ability to reorder pages using the thumbnail panel
+ * @property {string} PageNavigation Ability to navigate through pages using mouse wheel, arrow up/down keys and the swipe gesture.
```
achenPDFTron: im guessing we left out page up / page down on purpose?

ZhijieZhang:
> seeing `hotkeysManager` is camel cased, should we do the same with `TouchEventManager` to make it consistent? 

Hmm `TouchEventManager` is also camel cased. I guess you mean the first character should be lowercased? yeah I think we should make the naming consistent. I've added a card to my trello board and will do it in another PR.

ZhijieZhang:
> i think we need to add in page up and page down as well? as i can still navigate the page? 

ohhh good cacth, I missed these two keys. I will add them, thanks!

PR #501, opened 2020-02-07
[Bug] Link Button Sizing

andreysaf:
- Fixed the link button sizing for different languages
- Disable linking for comments since comments resize


ZhijieZhang:
If the fixes should be in 6.1 as well please make a same PR that points to 6.1. Thanks!

PR #502, opened 2020-02-07
[Bug] Fixed translations

andreysaf:
- Fixed display author vs author
- Fixed translations

### DIFF
--- a/src/components/NotePopup/NotePopup.js
+++ b/src/components/NotePopup/NotePopup.js
@@ -79,7 +79,7 @@ const NotePopup = ({ annotation, setIsEditing }) => {
   };
 
   const handleStateUpdate = state => {
-    const author = core.getDisplayAuthor(annotation);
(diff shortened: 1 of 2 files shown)


src/components/NotePopup/NotePopup.js
```diff
@@ -79,7 +79,7 @@ const NotePopup = ({ annotation, setIsEditing }) => {
   };
 
   const handleStateUpdate = state => {
-    const author = core.getCurrentUser();
+    const author = core.getDisplayAuthor(annotation);
```
ZhijieZhang: Hmm just to confirm with you that this is what we want because `getDisplayAuthor` returns the author of the annotation but `getCurrentUser` returns the current annotation(log in) user. They might not be the same.

andreysaf: Ahh, good catch. How would one get the display author name of a current user?

PR #506, opened 2020-02-11
[CHANGE] Updated instance.downloadPdf to take more download options. This includes `flatten`, which indicates whether all the annotations in the download document should be flattened

ZhijieZhang:


src/apis/downloadPdf.js
```diff
@@ -32,7 +53,12 @@ export default store => (includeAnnotations = true) => {
     return;
   }
 
-  downloadPdf(store.dispatch, {
-    includeAnnotations,
-  });
+  // legacy interface: instance.downloadPdf(true);
```
achenPDFTron: should we remove this?

src/apis/downloadPdf.js
```diff
@@ -32,7 +53,12 @@ export default store => (includeAnnotations = true) => {
...
+  if (typeof options === 'boolean') {
```
achenPDFTron: we can pass in a boolean only? i thought options had to be an object?

src/apis/downloadPdf.js
achenPDFTron: oh wait. i think you are referring to the fact that we support the legacy API / interface as well?

src/apis/downloadPdf.js
```diff
@@ -1,16 +1,37 @@
 /**
  * Downloads the pdf document with or without annotations added by WebViewer UI.
  * @method WebViewerInstance#downloadPdf
- * @param {boolean} [includeAnnotations=true] Whether or not to include annotations added by WebViewer UI.
+ * @param {object} [options] An object that contains the following download options.
+ * @param {string} [options.filename] The filename of the downloaded document.
+ * @param {string} [options.downloadType='pdf'] The type to download the file as, by default this is "pdf". PDF and image files can only be downloaded as PDFs, but office files can be downloaded as "pdf" or as "office" if you want to get the original file without annotations.
```
achenPDFTron: `"office" if you want to get the original file without annotations.`
slightly misleading?

I think if we had a pdf file that already had annots and set it  download type to office, i think the annots are still in there...?

achenPDFTron:
@ZhijieZhang i think that is all my comments for now

src/apis/downloadPdf.js
ZhijieZhang: idk haha I just copied the document from `getFileData` over here. How do you think we should explain this? I don't really know how this works haha.

src/apis/downloadPdf.js
ZhijieZhang: Hmm we can pass in an object but this API also takes in a boolean argument for backwards compatibility.

src/apis/downloadPdf.js
achenPDFTron: ok i see. hopefully we can deprecate the legacy interface in the future so the code is slightly easier to manage.

src/apis/downloadPdf.js
achenPDFTron: nevermind i misread it.
if you put download type as 'office' for a pdf file, it will still have its original annots.
but if its an office file, it wont have annots. so this comment is correct.

PR #510, opened 2020-02-18
Adds disableLogs state and stops the debug info logs if it is set

lbittner-pdftron:
The state isn't actually used but I made it for future use. Relies on https://github.com/XodoDocs/webviewer/pull/864

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -69,7 +69,7 @@ if (window.CanvasRenderingContext2D) {
     fullAPIReady = loadScript('../core/pdf/PDFNet.js');
   }
 
-  if (state.advanced.disableLogs) {
+  if (getHashParams('disableLogs', false)) {
     window.CoreControls.disableLogs(true);
   }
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -69,6 +69,10 @@ if (window.CanvasRenderingContext2D) {
     fullAPIReady = loadScript('../core/pdf/PDFNet.js');
   }
 
+  if (state.advanced.disableLogs) {
+    window.CoreControls.disableLogs(true);
```
ZhijieZhang: Let's just do `window.CoreControls.disableLogs(getHashParams('disableLogs', false))` instead of storing `disableLogs` in redux since it's not really viewer related and we don't have any actions to change this state. The `exposedSelectors` is a bit misleading since we aren't exposing the selectors in this file... If we'd like to add a getter to the API then we need to change `apis/index.js`

PR #511, opened 2020-02-19
Feature/ellipse measurement tool

bollain:
QA instructions are here: https://github.com/XodoDocs/webviewer/pull/874

### DIFF
--- a/src/components/MeasurementOverlay/EllipseMeasurementOverlay.js
+++ b/src/components/MeasurementOverlay/EllipseMeasurementOverlay.js
@@ -18,11 +18,16 @@ function EllipseMeasurementOverlay(props) {
   const renderScaleRatio = () => `${scale[0][0]} ${scale[0][1]} = ${scale[1][0]} ${scale[1][1]}`;
 
   const computeRadius = () => {
-    const decimalPlaces = 2;
+    const decimalPlaces = getNumberOfDecimalPlaces(annotation);
     const factor = annotation.Measure.axis[0].factor;
     const radiusInPts = (annotation.Width / 2).toFixed(decimalPlaces);
     return (radiusInPts * factor).toFixed(decimalPlaces);
   };
+
+  const getNumberOfDecimalPlaces = annotation =>
+    (annotation.Precision === 1
+      ? 0
+      : annotation.Precision.toString().split('.')[1].length);
 
   const onChangeRadiusLength = event => {
     const radius = Math.abs(event.target.value);
--- a/src/components/MeasurementOverlay/EllipseMeasurementOverlay.js
+++ b/src/components/MeasurementOverlay/EllipseMeasurementOverlay.js
@@ -2,14 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
 import Icon from 'components/Icon';
+import core from 'core';
+import getClassName from 'helpers/getClassName';
 import { mapAnnotationToKey, getDataWithKey } from '../../constants/map';
 
 function EllipseMeasurementOverlay(props) {
-  const { t, annotation } = props;
+  const { t, annotation, isOpen } = props;
   const annotationKey = mapAnnotationToKey(annotation);
   const { icon } = getDataWithKey(annotationKey);
   const scale = annotation.Scale;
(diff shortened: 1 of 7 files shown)


src/components/MeasurementOverlay/EllipseMeasurementOverlay.js
```diff
@@ -0,0 +1,107 @@
...
+  const { icon } = getDataWithKey(annotationKey);
+  const scale = annotation.Scale;
+  const precision = annotation.Precision;
+  const unit = annotation.Scale[1][1];
+  const className = getClassName('Overlay MeasurementOverlay', { isOpen });
+
+  const renderScaleRatio = () => `${scale[0][0]} ${scale[0][1]} = ${scale[1][0]} ${scale[1][1]}`;
+
+  const computeRadius = () => {
+    const decimalPlaces = 2;
```
ZhijieZhang: Do we want to use the logic in `getNumberOfDecimalPlaces` instead of having this to be fixed at 2?

src/components/MeasurementOverlay/EllipseMeasurementOverlay.js
```diff
@@ -0,0 +1,107 @@
...
+
+    if (diameterInPts > maxDiameterInPts) {
+      annotation.setHeight(maxDiameterInPts);
+      annotation.setWidth(maxDiameterInPts);
+      forceEllipseRedraw();
+    }
+  };
+
+  if (!isOpen) {
+    ensureDiameterIsWithinBounds(annotation.getWidth());
```
ZhijieZhang: Do you think we should wrap it with an `useEffect(() => {...}, [isOpen])` to be safe? At first I thought this will trigger an infinite rerendering because `ensureDiameterIsWithinBounds ` will call `forceEllipseRedraw`, which will trigger an `annotationChanged` event, which will rerender this component, and `ensureDiameterIsWithinBounds` will be called again when the overlay is closed. But now I see `if (diameterInPts > maxDiameterInPts)` will be false the second time the component rerenders after the overlay is closed.

src/components/MeasurementOverlay/MeasurementOverlay.js
```diff
@@ -146,7 +147,7 @@ class MeasurementOverlay extends React.PureComponent {
     if (key === 'perimeterMeasurement' || key === 'areaMeasurement') {
       // for polyline and polygon, there's no useful information we can show if it has no vertices or only one vertex.
       showInfo = annotation.getPath().length > 1;
-    } else if (key === 'distanceMeasurement') {
+    } else if (key === 'distanceMeasurement' || key === 'ellipseMeasurement') {
       showInfo = true;
```
ZhijieZhang: I noticed an edge case where if I just move my mouse a tiny bit when creating the ellipse measurement annotation, I won't see the annotation on the canvas, but the overlay will show up with area being 0 and the overlay won't go away. I feel like we need to do some extra check here(for example, if the circle is small) instead of just checking the key.

src/redux/initialState.js
```diff
@@ -126,6 +126,7 @@ export default {
       AnnotationCreateDistanceMeasurement: { dataElement: 'distanceMeasurementToolButton', title: 'annotation.distanceMeasurement', img: 'ic_annotation_distance_black_24px', group: 'measurementTools', showColor: 'active' },
       AnnotationCreatePerimeterMeasurement: { dataElement: 'perimeterMeasurementToolButton', title: 'annotation.perimeterMeasurement', img: 'ic_annotation_perimeter_black_24px', group: 'measurementTools', showColor: 'active' },
       AnnotationCreateAreaMeasurement: { dataElement: 'areaMeasurementToolButton', title: 'annotation.areaMeasurement', img: 'ic_annotation_area_black_24px', group: 'measurementTools', showColor: 'active' },
+      AnnotationCreateEllipseMeasurement: { dataElement: 'ellipseMeasurementToolButton', title: 'annotation.ellipseMeasurement', img: 'ic_annotation_circle_black_24px', group: 'measurementTools', showColor: 'active' },
```
ZhijieZhang: Do you think we should use another icon for the new tool? Maybe the icon should have dashed border(like measure polyline), or it's a circle that's filled with lines(like measure polygon).

PR #512, opened 2020-02-19
[Bugfix] Keyboard page scrolling fixes

awy3:
I fix a few issues with page scrolling with the arrow keys and the pageup/pagedown buttons. Problems I fix were

- Up/Down arrow keys will stop working when switching pages if you are zoomed in 
- For non-continuous mode with 2 pages, the mouse scroll wheel will go up and down while the up and down arrow keys will switch between right and left

Note sure if we want:
Should page up/down key change one page at a time when displaying two pages? When the user is zoomed out it doesn't make sense to do one page at time since going from page 3 to 4 might not do anything (if they are both displayed). However when WebViewer is zoomed in, the page up/down key will go from the left to right which make sense but feel unnatural since most of the time page up/down only does vertical moment.

### DIFF
--- a/src/components/DocumentContainer/DocumentContainer.js
+++ b/src/components/DocumentContainer/DocumentContainer.js
@@ -202,7 +202,7 @@ class DocumentContainer extends React.PureComponent {
 
     return (
       <div className={className} ref={this.container} data-element="documentContainer" onScroll={this.handleScroll} onTransitionEnd={this.onTransitionEnd}>
-        <div className="document" ref={this.document}></div>
+        <div className="document" ref={this.document} tabIndex="-1"></div>
       </div>
     );
   }
--- a/src/components/DocumentContainer/DocumentContainer.js
+++ b/src/components/DocumentContainer/DocumentContainer.js
@@ -9,6 +9,7 @@ import loadDocument from 'helpers/loadDocument';
 import getNumberOfPagesToNavigate from 'helpers/getNumberOfPagesToNavigate';
 import touchEventManager from 'helpers/TouchEventManager';
 import getHashParams from 'helpers/getHashParams';
+import setCurrentPage from 'helpers/setCurrentPage';
 import { getMinZoomLevel, getMaxZoomLevel } from 'constants/zoomFactors';
 import actions from 'actions';
 import selectors from 'selectors';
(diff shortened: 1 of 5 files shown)


src/components/DocumentContainer/DocumentContainer.js
```diff
@@ -202,7 +202,7 @@ class DocumentContainer extends React.PureComponent {
 
     return (
       <div className={className} ref={this.container} data-element="documentContainer" onScroll={this.handleScroll} onTransitionEnd={this.onTransitionEnd}>
-        <div className="document" ref={this.document}></div>
+        <div className="document" ref={this.document} tabIndex="-1"></div>
```
ZhijieZhang: Why do we need a `tabIndex`?

src/helpers/hotkeysManager.js
```diff
@@ -259,8 +268,11 @@ WebViewer(...)
         const { scrollHeight, clientHeight } = scrollViewElement;
         const reachedTop = scrollViewElement.scrollTop === 0;
         const currPageNumber = core.getCurrentPage();
-        if (reachedTop && currPageNumber > 1) {
-          core.setCurrentPage(currPageNumber - 1);
+        if (!this.isContinuous() && reachedTop && currPageNumber > 1) {
```
ZhijieZhang: Do we actually need the `isContinuous` function? If it's in continuous mode then `reachedTop` will be false when the current page isn't the first.

src/components/DocumentContainer/DocumentContainer.js
awy3: When using the up/down arrow buttons, I was noticing that the document would randomly lose focus. So if you have two page showing

- in non continuous mode as soon as the pages change, the up and down buttons stop working till you click on the document again
- in continuous mode, after scrolling pass a page with the arrow keys, it'll randomly lose focus and but button stops working

The `tabIndex` allow the document element to be focus using JavaScript. It seem some other code (probably in our core) is now able to keep focus on the document after adding this. I was trying stuff like 
```
setTimeout(() => {element.focus()} ,0)
```
but it turned out adding `tabIndex` was all I needed. I'll test this a bit more in Firefox, Edge, and IE to make sure this fix works

src/helpers/hotkeysManager.js
awy3: I guess we don't need this. I'll remove it

PR #514, opened 2020-02-20
[Bugfix] fixes the issue for invalid regex in search

sisardor:
### Description

Fixes the issue for invalid regex in search

### To verify
Regex in Chrome 
```js
readerControl.searchTextFull('(?<=(PDF\\s))(\\w+)', { regex: true }); // Valid 
readerControl.searchTextFull('(?=(PDF\\s)(\\w+)', { regex: true }); // Invlaid
```

Regex in Firefox
```js
readerControl.searchTextFull('(?<=(PDF\\s))(\\w+)', { regex: true }); // Invalid
readerControl.searchTextFull('(?=(PDF\\s))(\\w+)', { regex: true }); // Valid
readerControl.searchTextFull('(?=(PDF\\s)(\\w+)', { regex: true }); // Invalid
```

Screen Shot 2020-02-20 at 9 13 07 AM
Screen Shot 2020-02-20 at 9 14 39 AM

### Related issues

Fixes https://trello.com/c/Sk4OW3BR/767-invalid-regex-in-search-throws-error-and-gets-stuck

### DIFF
--- a/src/components/SearchOverlay/SearchOverlay.js
+++ b/src/components/SearchOverlay/SearchOverlay.js
@@ -114,9 +114,8 @@ class SearchOverlay extends React.PureComponent {
   }
 
   searchFailed = error => {
-    const { setIsSearching, setNoResult, setSearchError } = this.props;
+    const { setIsSearching, setSearchError } = this.props;
     setIsSearching(false);
-    setNoResult(true);
     if (error && error.message) {
       setSearchError(error.message);
     }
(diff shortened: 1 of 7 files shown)


src/components/SearchOverlay/SearchOverlay.js
```diff
@@ -104,6 +105,22 @@ class SearchOverlay extends React.PureComponent {
     }
   }
 
+  componentDidMount() {
+    core.addEventListener('searchFailed', this.searchFailed);
```
ZhijieZhang: Do you think we can listen to this event in `SearchPanel`? If we do this then we don't need to have an extra state and action for `setSearchError`. We can instead having an internal error state in `SearchPanel`.

sisardor: For this one, I think I just followed how `isSearching` and `noResult` are set in `SearchOverlay` and it flows down to SearchPanel. I thought in the way we keep `SearchPanel` simpler component.

If you think it should be in `SearchPanel` in its own state, let me know I can change it.

PR #515, opened 2020-02-20
[Feature] New APIs for selecting thumbnails

awy3:
I added new APIs for selecting/unselecting thumbnails

### DIFF
--- a/src/apis/index.js
+++ b/src/apis/index.js
@@ -127,6 +127,8 @@ import setSignatureFonts from './setSignatureFonts';
 import disableReplyForAnnotations from './disableReplyForAnnotations';
 import getCustomData from './getCustomData';
 import setCustomMeasurementOverlayInfo from './setCustomMeasurementOverlayInfo';
+import selectThumbnailPageNumbers from './selectThumbnailPageNumbers';
+import unselectThumbnailPageNumbers from './unselectThumbnailPageNumbers';
 
 export default store => {
   window.readerControl = {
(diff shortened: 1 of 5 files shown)


src/apis/selectThumbnailPageNumbers.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Select thumbnails in the thumbnail panel. This requires the "ThumbnailMultiselect" feature to be enabled
+ * @method WebViewer#selectThumbnailPageNumbers
```
ZhijieZhang: This should be `WebViewerInstance`. Are there any other apis that use `WebViewer`? We should change those as well.

src/apis/unselectThumbnailPageNumbers.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Unselect selected thumbnails
+ * @method WebViewer#unselelctThumbnailPageNumbers
```
ZhijieZhang: Should be WebViewerInstance.

src/apis/index.js
```diff
@@ -200,6 +202,8 @@ export default store => {
     setSignatureFonts: setSignatureFonts(store),
     setSelectedTab: setSelectedTab(store),
     getSelectedThumbnailPageNumbers: getSelectedThumbnailPageNumbers(store),
+    selectThumbnailPageNumbers: selectThumbnailPageNumbers(store),
```
ZhijieZhang: Personally I feel like `selectThumbnailPages` is better than what we have now. Ending the name with `Numbers` is a bit unnatural when I try to read it.

src/apis/selectThumbnailPageNumbers.js
awy3: Thanks for catching this, I'm not sure where the `WebViewer came from

src/apis/unselectThumbnailPageNumbers.js
awy3: Thanks for catching this, this has been fixed

src/apis/index.js
awy3: Since we had `getSelectedThumbnailPageNumbers` I was thinking of keeping the same naming but I can change it. Shorter name are better

PR #524, opened 2020-02-26
Bug fix/pop up api bugs

achenPDFTron:
https://trello.com/c/J0avN9BX/776-popup-api-bugs

### DIFF
--- a/src/components/AnnotationPopup/AnnotationPopup.js
+++ b/src/components/AnnotationPopup/AnnotationPopup.js
@@ -63,7 +63,7 @@ const AnnotationPopup = () => {
     // calling this function will always rerender this component
     // because the position state always has a new object reference
     const setPopupPositionAndShow = () => {
-      if (popupRef.current && popupItems && popupItems.length > 0) {
+      if (popupRef.current && popupItems.length > 0) {
         setPosition(
           getAnnotationPopupPositionBasedOn(firstAnnotation, popupRef),
         );
--- a/src/components/AnnotationPopup/AnnotationPopup.js
+++ b/src/components/AnnotationPopup/AnnotationPopup.js
@@ -25,6 +25,7 @@ const AnnotationPopup = () => {
     // can probably use mutation observer
     isLeftPanelOpen,
     isRightPanelOpen,
+    popupItems,
   ] = useSelector(
     state => [
       selectors.isElementDisabled(state, 'annotationPopup'),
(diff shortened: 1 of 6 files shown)


src/components/AnnotationPopup/AnnotationPopup.js
```diff
@@ -61,11 +63,10 @@ const AnnotationPopup = () => {
     // calling this function will always rerender this component
     // because the position state always has a new object reference
     const setPopupPositionAndShow = () => {
-      if (popupRef.current) {
+      if (popupRef.current && popupItems && popupItems.length > 0) {
```
achenPDFTron: prevent the "open" action from being dispatched because with if there are no items, it will show nothing in the DOM. so it might not make sense for it to have the state of "open"

src/components/AnnotationPopup/AnnotationPopup.scss
```diff
@@ -4,10 +4,6 @@
 .AnnotationPopup {
   @extend %popup;
 
-  &:empty {
```
achenPDFTron: moved this style into the popup.scss as the 3 popups need to use this style

src/constants/popup.scss
```diff
@@ -1,7 +1,6 @@
 @import './styles';
 
 $popup-button-height: 38px;
-$popup-button-width: 38px;
```
achenPDFTron: removed width because if there is a label associated with the popup button. it looks weird.

src/helpers/createPopupAPI.js
```diff
@@ -69,6 +69,9 @@ WebViewer(...)
   });
    */
   update(items) {
+    if (!items) {
+      items = [];
```
ZhijieZhang: Can we use an optional argument here?

src/components/TextPopup/TextPopup.js
```diff
@@ -40,15 +40,15 @@ const TextPopup = () => {
   useEffect(() => {
     const textSelectTool = core.getTool('TextSelect');
     const onSelectionComplete = (startQuad, allQuads) => {
-      if (popupRef.current) {
+      if (popupRef.current && popupItems && popupItems.length > 0) {
```
ZhijieZhang: Do we actually need `&& popupItems &&`?

achenPDFTron: Probably not because it looks like I'm getting `popupItems` from a selector and it looks like our Redux state initialize it to be [ ].

so it might be safe to just do `popupItems?.length > 0 ` ?

ZhijieZhang: If it’s always at least an empty array then I think we can access ‘length’ directly.

src/helpers/createPopupAPI.js
achenPDFTron: done

src/components/TextPopup/TextPopup.js
achenPDFTron: done

PR #525, opened 2020-02-26
[Refactor] Updates textSearchInit to use new function signature

sisardor:


src/components/SearchOverlay/SearchOverlay.js
```diff
@@ -164,8 +156,18 @@ class SearchOverlay extends React.PureComponent {
...
+
     setIsSearching(true);
-    core.textSearchInit(searchValue, searchMode, isFullSearch, handleSearchResult);
+
+    const options = {
+      'fullSearch': isFullSearch,
+      'onResult': handleSearchResult,
+      'onPageEnd': handleSearchResult,
+      'onDocumentEnd': handleSearchResult,
+      'onError': this.handleSearchError.bind(this),
```
ZhijieZhang: Why do we need to use `bind` here?

PR #527, opened 2020-02-29
Exposes setSelectedPageThumbnails for Feature.ThumbnailMultiselect

jonjanisch:
The selected thumbnails needs to be corrected after pages are moved or removed. ThumbnailsPanel is doing this internally, but that action is not exposed for external users of the WebViewer React app.

### DIFF
--- a/src/apis/index.js
+++ b/src/apis/index.js
@@ -101,6 +101,7 @@ import setPageLabels from './setPageLabels';
 import setPrintQuality from './setPrintQuality';
 import setReadOnly from './setReadOnly';
 import setSelectedTab from './setSelectedTab';
+import setSelectedThumbnailPageNumbers from './setSelectedThumbnailPageNumbers';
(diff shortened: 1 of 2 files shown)


awy3:
Thank you for the pull request and for your feedback. We actually had another pull request that has something similar to this (#515) that should be in our latest 6.2 release. We actually had the API name as `setSelectedThumbnailPageNumbers` like in your PR to match `getSelectedThumbnailPageNumbers` but decided to shorten it to [selectThumbnailPages](https://www.pdftron.com/api/web/WebViewerInstance.html#selectThumbnailPages__anchor).

Also if you get the latest copy of our master branch, there was a pull request (#529) which update the selected thumbnails after page operations. Lastly we have some more changes to the thumbnail panel in the future for you to look forward to, and any additional feedback would be appreciated. Thanks again for the feedback and the PR.

PR #528, opened 2020-03-03
[AEC] Annotation state

lbittner-pdftron:
Calls the view state API if constructor option is set.

https://github.com/XodoDocs/webviewer/pull/885


lbittner-pdftron:
Relies on https://github.com/XodoDocs/webviewer/pull/885 but should still work without it

PR #530, opened 2020-03-04
Bug fix/note panel styling

achenPDFTron:
Before:
image

After:
image
Note that there is a hover effect on hover of the drop down menu items

### DIFF
--- a/src/components/Dropdown/Dropdown.js
+++ b/src/components/Dropdown/Dropdown.js
@@ -6,6 +6,8 @@ import { withTranslation } from 'react-i18next';
 import actions from 'actions';
 import selectors from 'selectors';
 
+import Icon from 'components/Icon';
+
 import './Dropdown.scss';
 
 class Dropdown extends React.PureComponent {
@@ -60,7 +62,10 @@ class Dropdown extends React.PureComponent {
     return (
       <div className="Dropdown" data-element="dropdown" onClick={this.toggleDropdown}>
         <div className="items">
-          <div className="display-item">{this.getTranslatedContent(sortStrategy)}</div>
+          <div className="display-item">
+            <span>{this.getTranslatedContent(sortStrategy)}</span>
+            <Icon className="down-arrow" glyph="ic_arrow_drop_down_black_24px" />
+          </div>
           <div className={`dropdown-items ${this.state.isOpen ? 'show' : 'hide'}`}>
             {this.renderDropdownItems()}
           </div>
--- a/src/components/Dropdown/Dropdown.scss
+++ b/src/components/Dropdown/Dropdown.scss
@@ -48,10 +48,6 @@
         width: 100%;
         text-align: left;
         padding: 5px 5px;
-        text-overflow: ellipsis;
-        max-width: 100%;
-        white-space: nowrap;
-        overflow: hidden; 
         // need active so that it works on android mobile browser
         &:hover, &:active {
           background-color: rgba(0,0,0,.078);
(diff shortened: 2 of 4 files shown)


src/components/Dropdown/Dropdown.js
```diff
@@ -60,7 +62,10 @@ class Dropdown extends React.PureComponent {
     return (
       <div className="Dropdown" data-element="dropdown" onClick={this.toggleDropdown}>
         <div className="items">
-          <div className="display-item">{this.getTranslatedContent(sortStrategy)}</div>
+          <div className="display-item">
+            <span>{this.getTranslatedContent(sortStrategy)}</span>
+            <Icon className="down-arrow" glyph="ic_arrow_drop_down_black_24px" />
```
achenPDFTron: needed to add in icon so that the dropdown menu item width can stretch as long as the selected option.
Previously in the DOM, that wasn't the case. we just can't see it because we chose to have no border.

previously, the icon was being applied in the :after pseudo-selector

src/components/Dropdown/Dropdown.scss
```diff
@@ -7,8 +7,10 @@
   font-size: 14px;
   user-select: none;
   justify-content: space-between;
-  margin-bottom: 5px;
   z-index: 80;
+  border: 1px solid #e0e0e0;
+  border-radius: 4px;
+  background-color: white;
```
achenPDFTron: made border-radius 4 so that it matches the search field's border radius.

src/components/Dropdown/Dropdown.scss
```diff
@@ -18,8 +20,14 @@
       height: 24px;
       flex-grow: 1;
       display: flex;
-      justify-content: flex-end;
+      justify-content: space-between;
```
achenPDFTron: now that the drop down icon is an actual DOM element, needed to apply spacing between selected option and the icon else it looks ugly and bunched up.

src/components/Dropdown/Dropdown.scss
```diff
@@ -32,14 +35,27 @@
         display: flex;
         flex-direction: column;
         align-items: flex-end;
+
+        border: 1px solid #e0e0e0;
+        background-color: white;
```
achenPDFTron: styled the drop down items

src/components/Dropdown/Dropdown.scss
```diff
@@ -32,14 +35,23 @@
...
       }
   
       &.hide {
         display: none;
       }
 
       .dropdown-item {
-        margin: 5px 0;
```
achenPDFTron: removed margin and added it to padding so that the "hoverable area" looks bigger

src/components/Dropdown/Dropdown.scss
```diff
@@ -48,18 +60,6 @@
...
-    width: 24px;
-    height: 24px;
-
-    display: flex;
-    align-items: center;
-
-    content: '\25BE';
-    color: #757575;
-    color: var(--icon-color);
-  }
```
achenPDFTron: removed icon and made it an actual DOM element instead. see comment above for reasoning.

src/components/NotesPanel/NotesPanel.scss
```diff
@@ -14,12 +14,14 @@
...
 
+    margin: 0 16px;
+
     input {
       width: 150px;
-      margin-left: 15px;
       outline: none;
       border: 1px solid $gray12;
       border-radius: 4px;
+      margin-right: 5px;
```
achenPDFTron: added spacing between search input and dropdown

src/components/NotesPanel/NotesPanel.scss
```diff
@@ -14,12 +14,14 @@
     display: flex;
     justify-content: space-between;
 
+    margin: 0 16px;
```
achenPDFTron: added margin so that these 2 are aligned:
image

PR #531, opened 2020-03-05
Bugfix/rubber stamp overlay ie11

achenPDFTron:


src/components/StampOverlay/StampOverlay.scss
```diff
@@ -30,9 +30,6 @@
 
 
 .default-stamp-container {
-  display: flex;
-  flex-direction: column;
-
```
achenPDFTron: apparently when using ```flex-direction: column;```, in IE11, you need to specify the width as well ( https://stackoverflow.com/questions/42764591/flex-direction-column-makes-flex-items-overlap-ie11 )
https://stackoverflow.com/questions/37905749/flexbox-not-working-in-internet-explorer-11/42670221

However, observe:
image

There is only 1 element to apply flexbox on, I find it a bit redundant to apply flexbox on 1 element unless it greatly helps with positioning.

I removed the style and checked on all browers, and on an android phone and apple tablet.
Things work fine.
I checked with Sardor about this, it looks to be fine.

PR #535, opened 2020-03-13
[Feature] Moving pages between instance of WebViewer

awy3:
This PR is for moving pages between instance of WebViewer


awy3:
Became #559

PR #542, opened 2020-03-18
[Bugfix] Allow tooltip to work for customElement headers

awy3:
This fix allows tooltip to work for customElements. So you can do something like 
```
Webviewer({}, document.getElementById('viewer')).then(instance => {
  var checkboxLabelElement = {
    type: 'customElement',
    title: 'Tooltip popup message',
    render: function() {
      const label = document.createElement('label');
      label.innerHTML = 'Custom Label';
      return label;
    },
  };

  instance.setHeaderItems(function(header) {
    var items = header.getItems();
    items.splice(items.length - 2, 0, checkboxLabelElement);
    header.update(items);
  });
/*
```

and you'll get the tooltip when you hover over the element


awy3:
This was suppose to be for master not 6.2

PR #543, opened 2020-03-18
[Bugfix] Allow tooltip to work for customElement headers

awy3:
This fix allows tooltip to work for customElements. So you can do something like 
```
Webviewer({}, document.getElementById('viewer')).then(instance => {
  var checkboxLabelElement = {
    type: 'customElement',
    title: 'Tooltip popup message',
    render: function() {
      const label = document.createElement('label');
      label.innerHTML = 'Custom Label';
      return label;
    },
  };

  instance.setHeaderItems(function(header) {
    var items = header.getItems();
    items.splice(items.length - 2, 0, checkboxLabelElement);
    header.update(items);
  });
/*
```

and you'll get the tooltip when you hover over the element. 

Also for "intialState.js" we can add a title for "zoomOverlayButton" and maybe clean up "signatureToolButton" (SignatureToolButton.js is adding the tooltip for signature).

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect, useLayoutEffect } from 'react';
+import React, { useState, useRef, useEffect, useLayoutEffect, forwardRef } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -12,9 +12,24 @@ const propTypes = {
...
+    forwardedRef.current = targetRef.current;
+  }, [forwardedRef]);
+
+  return targetRef;
+}
+
+const Tooltip = forwardRef( ({ content = '', children }, forwardedRef) => {
   const timeoutRef = useRef(null);
-  const childRef = useRef(null);
+  const childRef = forwardedRef ? useCombinedRefs(forwardedRef) : useRef(null);
```
ZhijieZhang: Can you explain more about why `useCombinedRefs` is necessary here? I tried to remove this, and just having

```javascript
const childRef = forwardedRef ? forwardedRef : useRef(null);
```

also works.

awy3: Thank you for pointing this out. I was using an example I saw online but I guess we didn't need it after all

PR #544, opened 2020-03-19
[Bugfix] Hide links when other annotations are disabled

awy3:
This PR fix an issue where the "link" button is appearing when annotations are disabled. 

image

You can test with:
`readerControl.setReadOnly(true)`
`readerControl.disableAnnotations()`


awy3:
Should be merged into master......

PR #546, opened 2020-03-19
Fix issue with annotations still appearing in downloaded PDF

mparizeau-pdftron:
When includeAnnotations was false the original annotations
still appeared. It appears we need to pass an empty xfdf
string to make sure that no annotations show up.
Also made it to so that exportAnnotations isn't called
in this case or an overriding xfdfString is passed


mparizeau-pdftron:
I'd recommend hiding whitespace changes when looking at the diff

PR #548, opened 2020-03-23
[CHANGE] Fixed regresstion where switching between the thumbnail panel and another panel will result in warnings in the console.

ZhijieZhang:


awy3:
The changes seem good. However when I was testing this, I still see some warnings

ThumbnailWarning

It's not as bad as before but still show up when moving between pages really fast

ZhijieZhang:
Hi Andrew, thanks for the feedback. I will take a look at the remaining issue when I have time.

awy3:
You can merge it if you want, it improve the current problem.

ZhijieZhang:
@awy3 I looked into this a bit and it seems that the error is coming from WebViewer rendering pages, not thumbnails, since just scrolling the thumbnail panels won't output this warning message.

I will merge the changes in. Thanks for reviewing it.

PR #553, opened 2020-03-28
Features/disable style popup elements individually

achenPDFTron:


src/components/AnnotationStylePopup/AnnotationStylePopup.js
```diff
@@ -66,7 +66,9 @@ class AnnotationStylePopup extends React.Component {
           style={style}
           isFreeText={isFreeText}
           onStyleChange={this.handleStyleChange}
-          hideSlider={hideSlider}
+          hideOpacitySlider={hideSlider}
```
achenPDFTron: added extra prop so we have a bit more control over what kind of slider to hide.
and inside style pop up:
hideSlider = hideOpacitySlider && .....

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -9,6 +9,11 @@ import selectors from 'selectors';
 
 import './MeasurementOption.scss';
 
+const DATA_ELEMENTS = {
+  SCALE_INPUT_CONTAINER: 'scaleInputContainer',
```
achenPDFTron: added constants so its easier to work with rather than have hard coded strings everywhere

src/components/MeasurementOption/MeasurementOption.js
```diff
@@ -155,9 +162,14 @@ class MeasurementOption extends React.Component {
       { value: 0.0001, name: '0.0001' },
     ];
 
+    if (hideScaleInputContainer && hidePrecisionInputContainer) {
+      return null;
```
achenPDFTron: added this as we want a DOM element with height of 0 else it appears in DOM and looks weird

src/components/StylePopup/StylePopup.js
```diff
@@ -13,15 +13,24 @@ import selectors from 'selectors';
 
 import './StylePopup.scss';
 
+const DATA_ELEMENTS = {
```
ZhijieZhang, line 16: Can we name this `dataElements` instead, to match the conventions we have in WebViewer?

src/components/StylePopup/StylePopup.js
```diff
@@ -134,7 +162,10 @@ class StylePopup extends React.PureComponent {
 
 const mapStateToProps = (state, { colorMapKey }) => ({
   currentPalette: selectors.getCurrentPalette(state, colorMapKey),
-  isColorPaletteDisabled: selectors.isElementDisabled(state, 'colorPalette')
+  hideColorPalette: selectors.isElementDisabled(state, DATA_ELEMENTS.COLOR_PALETTE),
+  hideOpacitySlider: selectors.isElementDisabled(state, DATA_ELEMENTS.OPACITY_SLIDER),
+  hideStrokeThicknessSlider: selectors.isElementDisabled(state, DATA_ELEMENTS.STROKE_THICKNESS_SLIDER),
+  hideFontSizeSlider: selectors.isElementDisabled(state, DATA_ELEMENTS.FONT_SIZE_SLIDER),
```
ZhijieZhang, line 168: Can we use `isSomethingDisabled` as the variable name to match the naming convention in the other components?

src/components/ToolStylePopup/ToolStylePopup.js
```diff
@@ -139,7 +139,9 @@ class ToolStylePopup extends React.PureComponent {
           colorMapKey={colorMapKey}
           style={activeToolStyle}
           isFreeText={isFreeText}
-          hideSlider={hideSlider}
+          hideOpacitySlider={hideSlider}
```
ZhijieZhang, line 142: Hmm I see that in the `StylePopup` we are injecting props with the same names through redux, but here we are also passing them down. Why are we doing this?

I noticed that for the redaction tool and annotations, the sliders will show up now, which isn't what we want based on the code before.

I think to resolve this issue we can choose to not pass these props down, and inside `getAnnotationStyles.js` and `getToolStyles.js`, we can set the slider properties of redaction annotations or tool to null, so that sliders won't show up for redactions.

src/components/StylePopup/StylePopup.js
```diff
@@ -116,7 +144,7 @@ class StylePopup extends React.PureComponent {
           onMouseDown={e => e.preventDefault()}
         >
           <div className="sliders">
-            {!this.props.hideSlider && this.renderSliders()}
+            {!hideAllSlider && this.renderSliders()}
```
ZhijieZhang, line 147: Maybe have this conditional rendering around `<div className="sliders-container ..." >`, just to match your other code?

src/components/StylePopup/StylePopup.js
achenPDFTron: but these are just constants . shouldn't they be in upper case?

src/components/StylePopup/StylePopup.js
achenPDFTron: addressed

src/components/StylePopup/StylePopup.js
achenPDFTron: addressed

src/components/ToolStylePopup/ToolStylePopup.js
achenPDFTron: addressed

achenPDFTron: seems i was not aware, that `hideSlider` was not injected as a prop through redux. that was a bug on my part.

src/components/StylePopup/StylePopup.js
ZhijieZhang: Oh my bad, it should be `DataElements`. I'm not saying keep them all uppercase is wrong or something. Just want to make sure that for new constants they should match the naming convention we have in WebViewer.

achenPDFTron: changed.
Seems I forgot that we agreed on a convention for constants a while.

PR #558, opened 2020-04-03
[CHANGE] can style grouped annotation

achenPDFTron:
We can style grouped annotations.
It doesn't quite match Adobe's, but it should be intuitive enough.

Assuming an annotation is grouped, if you click on it and style it, it will only style the annotation you clicked on.

This feature was requested by a customer.

If we go Adobe's way, it may take a bit more time with implementation


achenPDFTron:
Note to self, going to let customer see it first for feedback

PR #559, opened 2020-04-06
[Feature] Moving pages between instance of WebViewer

awy3:
This PR is for moving pages between instance of WebViewer


ZhijieZhang:
Andrew has made a new PR for this feature, so I will close this one.

PR #572, opened 2020-04-16
[CHANGE] fix(AnnotationContentOverlay): do not show the component when a modal is open

### LINKED ISSUE #570: Highlight hover effect exists while creating Signature
If you highlight text and then attempt to create a signature then the hover effect still fires.  Users have pointed out that it's distracting.

Note, may exist in other modal/popup situations.

image

ZhijieZhang:
Checks if the mouse in moving on a modal component to decide if we are hovering on an annotation. Fixes #570.

### DIFF
--- a/src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
+++ b/src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
@@ -22,26 +22,13 @@ const AnnotationContentOverlay = () => {
   });
 
   useEffect(() => {
-    const isMouseOnModal = e => {
-      let node = e.target;
-
-      while (node && node !== document.body) {
-        if (node.classList.contains('Modal')) {
-          return true;
-        }
-
-        node = node.parentNode;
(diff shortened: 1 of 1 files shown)


ZhijieZhang:
Close for a while as I noticed there're more components we need to take into account. Will reopen once it's fixed.

ZhijieZhang:
@achenPDFTron Can you review this when you have time? Thanks.

achenPDFTron:
Note to self, able to replicate the bug on https://www.pdftron.com/samples/web/samples/viewing/viewing/

achenPDFTron:
@ZhijieZhang 
When I have the redaction modal opened, I can still hover over the redaction annot I have created. Is that intentional?

ZhijieZhang:
I can reproduce the issue where the redaction annotation can still be hovered when the modal is open, but the comment text won't show up anymore.

I guess if we want to fix the redaction hovering issue we'd want it in another PR.

PR #573, opened 2020-04-16
[CHANGE] Able to bring up color palette to style selected custom crea…

achenPDFTron:
…ted annotations

### DIFF
--- a/src/apis/registerTool.js
+++ b/src/apis/registerTool.js
@@ -9,6 +9,7 @@
  * @param {string} [properties.buttonGroup] Group of the tool button belongs to.
  * @param {string} [properties.tooltip] Tooltip of the tool button.
  * @param {function} [annotationConstructor] The constructor function for the annotation that will be created by the registered tool.
+ * @param {function} [customAnnotationCheckFxn] Function that takes in a parameter of an annotation. Returns a boolean if the specified annotation is a certain type of annotation. When this is defined, annotationConstructor can be omitted. Must define it when creating a custom annotation
  * @example
 WebViewer(...)
   .then(function(instance) {
@@ -22,7 +23,7 @@ WebViewer(...)
       tooltip: 'MyTooltip'
     };
 
-    instance.registerTool(myTool, myCustomAnnotation);
+    instance.registerTool(myTool, undefined, annot => annot && annot.isCustomAnnot);
   });
  */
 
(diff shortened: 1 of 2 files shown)


src/apis/registerTool.js
```diff
@@ -22,18 +23,18 @@ WebViewer(...)
...
+    instance.registerTool(myTool, undefined, annot => annot && annot.isCustomAnnot);
   });
  */
 
 import core from 'core';
 import { register, copyMapWithDataProperties } from 'constants/map';
 import actions from 'actions';
-
-export default store => (tool, annotationConstructor) => {
+// TODO may want to remove annotationConstructor in 6.4.0 as custoomANnotationCheckFxn makes it sort of redundant
```
ZhijieZhang, line 33: It should be `7.0` instead of `6.4.0`

src/apis/registerTool.js
```diff
@@ -9,6 +9,7 @@
  * @param {string} [properties.buttonGroup] Group of the tool button belongs to.
  * @param {string} [properties.tooltip] Tooltip of the tool button.
  * @param {function} [annotationConstructor] The constructor function for the annotation that will be created by the registered tool.
+ * @param {function} [customAnnotationCheckFxn] Function that takes in a parameter of an annotation. Returns a boolean if the specified annotation is a certain type of annotation. When this is defined, annotationConstructor can be omitted. Must define it when creating a custom annotation
```
ZhijieZhang, line 12: Is `Fxn` a widely used word? Can we use `Func` or `Function` instead?

Also, can we change this description to mention something like, `this function is used by the viewer to check if the annotation passed in is associated(created) with the registered tool`?

src/apis/registerTool.js
```diff
@@ -9,6 +9,7 @@
  * @param {string} [properties.buttonGroup] Group of the tool button belongs to.
  * @param {string} [properties.tooltip] Tooltip of the tool button.
  * @param {function} [annotationConstructor] The constructor function for the annotation that will be created by the registered tool.
```
ZhijieZhang, line 11: Can we change the description to this argument to say it's deprecated, and the third argument should be used? Thanks!

src/apis/registerTool.js
achenPDFTron: addressed

PR #574, opened 2020-04-16
Updates to behaviour on Enter key based on call

bollain:
Made some updates to the behaviour on Enter based on call with LCT.

Client wants the annotation to be finished on ENTER, and also wants the overlay to close. I did this by deselecting the annotation programatically; an alternative approach would be passing the `closeElement` callback to the components. I chose deselected because it triggers an annotationEvent; I will send a custom build to the client and see if this will work for them; they seem to rely on events for some other stuff they are doing.

Also forgot to add the ENTER onkey for the EllipseMeasurement, and Janice provided a shiny new icon.


src/constants/map.js
```diff
@@ -67,7 +67,7 @@ const map = {
       annotation.Measure,
   },
   rectangularAreaMeasurement: {
-    icon: 'ic_annotation_rectangular_area_black_24px',
+    icon: 'ic_annotation_ellipse_area_black',
```
ZhijieZhang, line 70: Hmm should we keep the original value? If we change this to `ellipse` then the icon in the notes panel of a rectangular polygon annotation will be changed to `ellipse`

src/constants/map.js
```diff
@@ -90,7 +90,7 @@ const map = {
       annotation.Measure,
   },
   ellipseMeasurement: {
-    icon: 'ic_annotation_circle_black_24px',
+    icon: '',
```
ZhijieZhang, line 93: I think we want this to be `ic_annotation_ellipse_area_black`?

bollain: Thanks for catching this @ZhijieZhang, I copy pasted on to the wrong place.

PR #577, opened 2020-04-21
feat: add an 'showColor 'option to instance.registerTool and toolGroupButton

ZhijieZhang:
This is the code that I used to test
```javascript
Webviewer(
  {
    initialDoc: hashFile,
    path: '/lib',
  },
  document.getElementById('viewer')
).then((instance) => {
  const { docViewer, Tools, Annotations } = instance;

  const groupName = 'customGroup';
  const customRectangleTool = new Tools.RectangleCreateTool(docViewer);
  const customPolygonTool = new Tools.RectangleCreateTool(docViewer);
  const customTools = [
    {
      toolName: 'CustomRectangleTool',
      toolObject: customRectangleTool,
      buttonImage: 'ic_annotation_square_black_24px',
      buttonName: 'customRectangleTool',
      buttonGroup: groupName,
      showColor: 'always',
    },
    {
      toolName: 'CustomPolygonTool',
      toolObject: customPolygonTool,
      buttonImage: 'ic_annotation_polygon_black_24px',
      buttonName: 'customPolygonTool',
      buttonGroup: groupName,
      showColor: 'always',
    },
  ];

  customTools.forEach((toolData) => {
    instance.registerTool(toolData);
  });

  // set some default colors for the custom tools.
  customRectangleTool.setStyles({
    StrokeColor: new Annotations.Color(255, 0, 0)
  })
  customPolygonTool.setStyles({
    StrokeColor: new Annotations.Color(255, 0, 0)
  })

  instance.setHeaderItems((header) => {
    header.get('freeHandToolGroupButton').insertBefore({
      type: 'toolGroupButton',
      toolGroup: groupName,
      dataElement: 'customToolGroupButton',
      showColor: 'always',
    });
  });
});
```

Using the code above, you should see that the colors of the tool group button and the buttons inside the group are always showing without activating them.


src/redux/reducers/viewerReducer.js
```diff
@@ -118,7 +118,7 @@ export default initialState => (state = initialState, action) => {
             title: payload.tooltip,
             group: payload.buttonGroup,
             img: payload.buttonImage,
-            showColor: 'active',
+            showColor: payload.showColor || 'active',
```
achenPDFTron, line 121: we are also using `active` in another changed file;  `if (showColor === 'always' || (showColor === 'active' && isActive)) {`.

should we have a constant for these?

ZhijieZhang: I think we will use more constants in this projects at a point, maybe after the new UI is merged.
For now I'm ok with having this just to be a string. It's not inconsistent with other code.

PR #579, opened 2020-04-21
feat(api): add setActivePalette and setColorPalette APIs

ZhijieZhang:
`setActivePalette` allows users to programmatically set the active color palette for a tool and its annotations.
`setColorPalette` allows users to set different colors for all of the tools or specific tools.

```javascript
instance.setColorPalette(['#FFFFFF', '#001133', '#DDDDDD', '#12345D']);
instance.setColorPalette({
  toolNames: ['AnnotationCreateRectangle', 'AnnotationCreatePolygon'],
  colors: ['#FFFFFF', 'transparency', '#000000'],
});
instance.setColorPalette({
  toolNames: ['AnnotationCreateFreeText'],
  colors: ['#000000'],
});
```

### DIFF
--- a/src/apis/setColorPalette.js
+++ b/src/apis/setColorPalette.js
@@ -1,14 +1,61 @@
+import { mapToolNameToKey } from 'constants/map';
 import actions from 'actions';
-import { getDataWithKey, mapToolNameToKey } from 'constants/map';
-import mapPaletteToAnnotationColorProperty from 'constants/mapPaletteToAnnotationColorProperty';
-import mapAnnotationColorPropertyToPalette from 'constants/mapAnnotationColorPropertyToPalette';
+import selectors from 'selectors';
 
-export default store => (toolName, colorPalette) => {
-  const availablePalettes = getDataWithKey(mapToolNameToKey(toolName)).availablePalettes;
-  const property = mapPaletteToAnnotationColorProperty[colorPalette];
-  if (availablePalettes.includes(property)) {
-    store.dispatch(actions.setColorPalette(mapToolNameToKey(toolName), property));
-  } else {
-    console.warn(`${toolName} does not have ${colorPalette} color, available colors are: ${availablePalettes.map(palette => mapAnnotationColorPropertyToPalette[palette]).join(', ')} `);
+/**
+ * @typedef WebViewerInstance.PaletteOption
+ * @property {string[]} toolNames Tools that will have the same colors in the palette.
+ * @property {string[]} colors An array of hex colors. Use 'transparency' for a transparent color.
+ */
+
+/**
+ * Sets the colors in the palette globally or for specific tools and their associated annotations
+ * @method WebViewerInstance#setColorPalette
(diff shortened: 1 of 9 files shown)


src/apis/setColorPalette.js
```diff
@@ -1,14 +1,61 @@
...
+/**
+ * @typedef WebViewerInstance.PaletteOption
+ * @property {string[]} toolNames Tools that will have the same colors in the palette.
+ * @property {string[]} colors An array of hex colors. Use 'transparency' for a transparent color.
+ */
+
+/**
+ * Sets the colors in the palette globally or for specific tools and their associated annotations
+ * @method WebViewerInstance#setColorPalette
+ * @param {string[]|WebViewerInstance.PaletteOption} An array of hex colors that will override the default colors for every tool.
```
achenPDFTron, line 14: are you sure you want to have 2 different kinds of params?
Won't it be a little bit confusing?

What are your thoughts on:
fxn( toolsArray, colorsArray)

?

achenPDFTron:
@ZhijieZhang 
```
    instance.setColorPalette({
      // toolNames: ['AnnotationCreateRectangle', 'AnnotationCreatePolygon'],
      colors: ['#FFFFFF', 'transparency', '#000000'],
    });
```
The above does not seem to do anything....

Is that intetional?

achenPDFTron:
@ZhijieZhang i played around with
`instance.setActivePalette('AnnotationCreateFreeText', 'fill')`

I don't quite see what it does?

ZhijieZhang:
> @ZhijieZhang
> 
> ```
>     instance.setColorPalette({
>       // toolNames: ['AnnotationCreateRectangle', 'AnnotationCreatePolygon'],
>       colors: ['#FFFFFF', 'transparency', '#000000'],
>     });
> ```
> 
> The above does not seem to do anything....
> 
> Is that intetional?

Yep according to the documentation this is not a valid argument.

ZhijieZhang:
> @ZhijieZhang i played around with
> `instance.setActivePalette('AnnotationCreateFreeText', 'fill')`
> 
> I don't quite see what it does?

Load WebViewer and click on the freetext tool button, I think by default it shows the `text` color palette. If you use the code above you should see the palette is changed to `fill`.

src/apis/setColorPalette.js
ZhijieZhang: I think I'm ok with the current signature. We have some examples in WebViewer that does this. For example: `loadCanvasAsync`.

Change the signature to be `toolsArray, colorsArray` makes it harder to document. Since the first argument will be `string[]` in both cases, imo it's not very clear to see if it's actually a tool names array or a colors array.

We could make it be `colorsArray, toolsArray`, but I think making it as an object is more explicit.

I think in the future we may use objects as an argument anyways when there's a request for having different colors for different palettes. In this case we could change it to:
```javascript
setColorPalette({
  toolNames: ['...'],
  text: ['...'],
  fill: ['...'],
  border: ['...'],
})
```

PR #581, opened 2020-04-22
Bug fix/custom annot check fxn

achenPDFTron:
https://github.com/PDFTron/webviewer-ui/pull/573


src/constants/map.js
```diff
@@ -345,9 +345,9 @@ export const register = (tool, annotationConstructor) => {
     currentPalette: availablePalettes[0],
     availablePalettes,
     toolNames: [toolName],
-    annotationCheck: annotationConstructor
-      ? annotation => annotation instanceof annotationConstructor
-      : null,
+    annotationCheck: customAnnotCheckFunc ?
+      annotation => customAnnotCheckFunc(annotation) :
```
ZhijieZhang, line 349: I think in this case we can just return `customAnnotCheckFunc`, no need to create another function that just pass the argument to `customAnnotCheckFunc`. But not a big deal.

PR #582, opened 2020-04-22
[CHANGE] able to show thumbnails like a grid on drag of thumbnail panel

achenPDFTron:
When dragging the thumbnail panel, the thumbnails don't seem to be rendered like a grid when its dragged to its max width. it still stays as 1 column. Only when you resize the browser, it will happen.

Made change so that it can have 1-3 depending on the thumbnail width without resizing the browser.

Adobe seems to have this kind of feature.

### DIFF
--- a/src/components/ThumbnailsPanel/ThumbnailsPanel.js
+++ b/src/components/ThumbnailsPanel/ThumbnailsPanel.js
@@ -213,22 +213,6 @@ class ThumbnailsPanel extends React.PureComponent {
   }
 
   getNumberOfColumns = () => {
-    // const desktopBreakPoint = 640;
-    // const { innerWidth } = window;
-    // let numberOfColumns;
-
-    // if (innerWidth >= desktopBreakPoint) {
-    //   numberOfColumns = 1;
-    // // TODO: use forwardRef to get the width of the thumbnail div instead of using the magic 20px
-    // } else if (innerWidth >= 3 * (THUMBNAIL_SIZE + 20)) {
-    //   numberOfColumns = 3;
-    // } else if (innerWidth >= 2 * (THUMBNAIL_SIZE + 20)) {
-    //   numberOfColumns = 2;
-    // } else {
-    //   numberOfColumns = 1;
-    // }
-
-    // return numberOfColumns;
     const numberOfColumns = Math.min(3, Math.max(1, Math.floor(this.state && this.state.width ? this.state.width / 160 : 0)));
     return numberOfColumns;
   }
--- a/src/components/ThumbnailsPanel/ThumbnailsPanel.js
+++ b/src/components/ThumbnailsPanel/ThumbnailsPanel.js
@@ -35,7 +35,6 @@ class ThumbnailsPanel extends React.PureComponent {
     this.listRef = React.createRef();
     this.afterMovePageNumber = null;
     this.state = {
-      numberOfColumns: 1,
       isDocumentControlHidden: true,
       canLoad: true,
       height: 0,
@@ -53,7 +52,6 @@ class ThumbnailsPanel extends React.PureComponent {
     core.addEventListener('pageNumberUpdated', this.onPageNumberUpdated);
     core.addEventListener('pageComplete', this.onPageComplete);
     core.addEventListener('annotationHidden', this.onAnnotationChanged);
-    window.addEventListener('resize', this.onWindowResize);
   }
 
   componentWillUnmount() {
@@ -63,7 +61,6 @@ class ThumbnailsPanel extends React.PureComponent {
     core.removeEventListener('pageNumberUpdated', this.onPageNumberUpdated);
     core.removeEventListener('pageComplete', this.onPageComplete);
     core.removeEventListener('annotationHidden', this.onAnnotationChanged);
-    window.removeEventListener('resize', this.onWindowResize);
   }
 
   onBeginRendering = () => {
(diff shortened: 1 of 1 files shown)


src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -432,6 +434,10 @@ class ThumbnailsPanel extends React.PureComponent {
             this.setState({
               height: bounds.height,
               width: bounds.width,
+            }, () => {
+              this.setState({
```
ZhijieZhang, line 438: Can we make a `handleResize` function and calculate the number of columns there, and just setState once? I don't think it's necessary to rerender the component twice. (one for setting the dimension, one for setting the columns)

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -213,21 +213,23 @@ class ThumbnailsPanel extends React.PureComponent {
...
-    // TODO: use forwardRef to get the width of the thumbnail div instead of using the magic 20px
-    } else if (innerWidth >= 3 * (THUMBNAIL_SIZE + 20)) {
-      numberOfColumns = 3;
-    } else if (innerWidth >= 2 * (THUMBNAIL_SIZE + 20)) {
-      numberOfColumns = 2;
-    } else {
-      numberOfColumns = 1;
-    }
-
+    // const desktopBreakPoint = 640;
```
ZhijieZhang, line 216: Can we remove the commented code?

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -213,21 +213,23 @@ class ThumbnailsPanel extends React.PureComponent {
...
+    // } else if (innerWidth >= 3 * (THUMBNAIL_SIZE + 20)) {
+    //   numberOfColumns = 3;
+    // } else if (innerWidth >= 2 * (THUMBNAIL_SIZE + 20)) {
+    //   numberOfColumns = 2;
+    // } else {
+    //   numberOfColumns = 1;
+    // }
+
+    // return numberOfColumns;
+    const numberOfColumns = Math.min(3, Math.max(1, Math.floor(this.state && this.state.width ? this.state.width / 160 : 0)));
```
ZhijieZhang, line 232: it seems to me that `this.state.width` will always be defined? If this is the case can we remove the unnecessary check. Thanks!

src/components/ThumbnailsPanel/ThumbnailsPanel.js
achenPDFTron: done

src/components/ThumbnailsPanel/ThumbnailsPanel.js
achenPDFTron: done

src/components/ThumbnailsPanel/ThumbnailsPanel.js
achenPDFTron: done

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -35,7 +35,7 @@ class ThumbnailsPanel extends React.PureComponent {
     this.listRef = React.createRef();
     this.afterMovePageNumber = null;
     this.state = {
-      numberOfColumns: this.getNumberOfColumns(),
+      numberOfColumns: 1,
```
ZhijieZhang, line 38: Hmm after another look it seems to me that it's not necessary to store `numberOfColumns` in `this.state` anymore. Since we can just calculate this value from `width`.

Can we remove this from state? Sorry for not finding this out in the first place...

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -208,26 +208,12 @@ class ThumbnailsPanel extends React.PureComponent {
 
   onWindowResize = () => {
     this.setState({
-      numberOfColumns: this.getNumberOfColumns(),
+      numberOfColumns: this.getNumberOfColumns(this.state.width),
```
ZhijieZhang, line 211: I think we can remove this function now, as we will get columns from width

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -208,26 +208,12 @@ class ThumbnailsPanel extends React.PureComponent {
...
-    } else if (innerWidth >= 3 * (THUMBNAIL_SIZE + 20)) {
-      numberOfColumns = 3;
-    } else if (innerWidth >= 2 * (THUMBNAIL_SIZE + 20)) {
-      numberOfColumns = 2;
-    } else {
-      numberOfColumns = 1;
-    }
-
+  getNumberOfColumns = width => {
+    const numberOfColumns = Math.min(3, Math.max(1, Math.floor(width / 160)));
```
ZhijieZhang, line 216: Let's just have this in the `renderThumbnails` instead of a separate function.

achenPDFTron: as discussed on Xodo, lets keep this fxn so we don't have to rewrite the logic as we need to get the number of columns in several places

src/components/ThumbnailsPanel/ThumbnailsPanel.js
achenPDFTron: removed

src/components/ThumbnailsPanel/ThumbnailsPanel.js
achenPDFTron: removed the `onWindowResize` fxn

PR #583, opened 2020-04-23
[Feature] i18n support to datepicker widget

sisardor:


src/apis/setDatePickerLocale.js
```diff
@@ -0,0 +1,17 @@
+import core from 'core';
+
+export default function(i18nextPromise) {
+  i18nextPromise.then(t => {
```
ZhijieZhang, line 4: This file should be put under the `helpers` folder as this is not an API that's attached to `window.readerControl`.

PR #584, opened 2020-04-24
[Feature] Update page operation defaults

awy3:
This PR is to make reordering pages and merging documents enabled by default. However since merging pages between instance of WebViewer has a lot of overhead (it need to extract pages onDragStart), I added a `MultipleViewerMerging` feature


src/constants/feature.js
```diff
@@ -36,6 +36,7 @@ export default {
   TextSelection: 'TextSelection',
   TouchScrollLock: 'TouchScrollLock',
   Copy: 'Copy',
+  MultipleViewerMerging: 'MultipleViewerMerging',
```
ZhijieZhang, line 39: I think we need to have some documentation about this feature.

awy3: Thanks for the reminder. I added a comment to the top of this page

PR #585, opened 2020-04-27
[Feature] Merging document event

awy3:
A new event letting user know what being merged into the document


src/helpers/extractPagesWithAnnotations.js
```diff
@@ -20,7 +20,7 @@ export default async pagesToExtract => {
       annotManager.exportAnnotations({ annotList }).then(xfdfString => {
         doc.extractPages(pagesToExtract, xfdfString).then(data => {
           const arr = new Uint8Array(data);
-          const fileName = 'extractedDocument.pdf';
+          const fileName = doc.filename || 'extractedDocument.pdf';
```
ZhijieZhang, line 23: Is `filename` exposed in production build? Just want to double check with you.

src/helpers/pageManipulation.js
```diff
@@ -38,9 +45,11 @@ export const mergeExternalWebViewerDocument = (viewerID, mergeToPage) => dispatc
 
     dispatch(actions.openElement('loadingModal'));
     extractedDataPromise.then(docToMerge => {
-      dispatch(mergeDocument(docToMerge, mergeToPage)).then(pageInserted => {
+      dispatch(mergeDocument(docToMerge, mergeToPage, false)).then(({ fileName, pages }) => {
+
+        fireEvent('documentMerged', { fileName, pages: otherWebViewerIframe.contentWindow.pagesExtracted });
```
ZhijieZhang, line 50: Could you tell me why this function doesn't need a `shouldFireEvent` argument like `mergeDocument` requires(though it's optional)?

src/helpers/extractPagesWithAnnotations.js
awy3: It show up in `readerControl.docViewer.getDocument().filename` but I'm not sure what setting that or if it's reliable so I'll change it

src/helpers/pageManipulation.js
awy3: `mergeExternalWebViewerDocument` is getting the document to merge and using `mergeDocument` to merge it. Since the "pages" in `mergeDocument` isn't going to be correct (we want to know which pages of the original document were merged), I added `shouldFireEvent` to keep `mergeDocument` from firing the "documentMerged" event (`mergeExternalWebViewerDocument` will fire it with the correct value). 

I can add a `shouldFireEvent` option to `mergeExternalWebViewerDocument` but there doesn't seem to be any point since nothing going to be using it at the moment. Alternatively, I can also pass an "override  pages" option to "mergeDocument" to change data in the "documentMerged" event, but decided to go with a `shouldFireEvent` flag instead. Let me know which one you perfer

ZhijieZhang: Hmm I don't have a preference. I'm just curious about the reason.

PR #587, opened 2020-04-27
[Feature] Enable feature by default

awy3:
Enable thumbnail controls by default

image

### DIFF
--- a/src/components/ThumbnailsPanel/ThumbnailsPanel.js
+++ b/src/components/ThumbnailsPanel/ThumbnailsPanel.js
@@ -12,6 +12,8 @@ import { extractPagesToMerge, mergeExternalWebViewerDocument, mergeDocument } fr
 import core from 'core';
 import selectors from 'selectors';
 import actions from 'actions';
+import { workerTypes } from 'constants/types';
+import { PRIORITY_THREE, PRIORITY_ONE } from 'constants/actionPriority';
 
 import './ThumbnailsPanel.scss';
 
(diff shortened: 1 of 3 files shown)


src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -224,7 +226,15 @@ class ThumbnailsPanel extends React.PureComponent {
   };
 
   onDocumentLoaded = () => {
-    const { setSelectedPageThumbnails } = this.props;
+    const { setSelectedPageThumbnails, dispatch } = this.props;
+
+    let doc = core.getDocument();
+    if (doc.type !== workerTypes.PDF) {
+      dispatch(actions.disableElements(['thumbnailControl',], PRIORITY_ONE));
```
ZhijieZhang, line 233: Is the `,` inside the array intentional?

src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -224,7 +226,15 @@ class ThumbnailsPanel extends React.PureComponent {
   };
 
   onDocumentLoaded = () => {
-    const { setSelectedPageThumbnails } = this.props;
+    const { setSelectedPageThumbnails, dispatch } = this.props;
+
+    let doc = core.getDocument();
+    if (doc.type !== workerTypes.PDF) {
```
ZhijieZhang, line 232: If a user loads an office file, we will disable thumbnail operations, which makes sense to me. But if the user then loads a PDF document, do we want to re-enable the controls back?

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: Thanks for pointing this out. I actually remove the "disableElements" here

src/components/ThumbnailsPanel/ThumbnailsPanel.js
awy3: I changed it so if the controls are enabled and the user open a xod file the controls disappear. If they open a pdf after, as long as they didn't use "disableFeature" or something, the controls will be re-enabled. 

One thing to note is I added a "showShouldControls" flag on the state of the thumbnailsPanel instead of the global store. I could move it to the global store but I'm not sure how useful it would be at this point

PR #588, opened 2020-04-28
[Bugfix] Stop signature from entering into an error state 

awy3:
After the signature pop up show up, users can use hot keys to change the current tool. So it's possible for the signature modal to be displayed when the active tool has been change. If the user than click "create", there will be a signature preview even when the tool has already changed

EnteringErrorState

My fix disable changing tools when the signature overlay is opened. I don't think there are any other cases similar to this one (is there?)

### DIFF
--- a/src/helpers/hotkeysManager.js
+++ b/src/helpers/hotkeysManager.js
@@ -467,7 +467,7 @@ WebViewer(...)
       const openElements = selectors.getOpenElements(getState());
       const currentToolName = core.getToolMode().name;
     
-       // disable changing tool when the signature overlay is opened.
+      // disable changing tool when the signature overlay is opened.
       const isSignatureModalOpen = currentToolName === window.Tools.ToolNames.SIGNATURE && openElements['signatureModal'];
 
       if (isFocusingElement() || isSignatureModalOpen) {
--- a/src/helpers/hotkeysManager.js
+++ b/src/helpers/hotkeysManager.js
@@ -470,7 +470,7 @@ WebViewer(...)
        // disable changing tool when the signature overlay is opened.
       const isSignatureModalOpen = currentToolName === window.Tools.ToolNames.SIGNATURE && openElements['signatureModal'];
 
-      if (isFocusingElement() && isSignatureModalOpen) {
+      if (isFocusingElement() || isSignatureModalOpen) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


src/helpers/setToolModeAndGroup.js
```diff
@@ -5,6 +5,15 @@ import selectors from 'selectors';
 
 export default (store, toolName) => {
   const { dispatch, getState } = store;
+
+  const openElements = selectors.getOpenElements(getState());
```
ZhijieZhang, line 9: I think a better place to put this logic is at 
https://github.com/PDFTron/webviewer-ui/blob/9d85f21e0f9efe8febf74d6b71f06886ef200a62/src/helpers/hotkeysManager.js#L408-L419
I think having code in this above function is more intuitive to think. We want to disable tool hotkeys when the modal is open, not still allow hotkeys but make an exception for the modal in this function. What do you think?

awy3: I moved the code to the `createToolHotkeyHandler` function in `hotkeysManager`

src/helpers/hotkeysManager.js
```diff
@@ -460,7 +461,17 @@ WebViewer(...)
    * @ignore
    */
   createToolHotkeyHandler(handler) {
+    const { getState } = this.store;
+
     return (...args) => {
+      const openElements = selectors.getOpenElements(getState());
+      const currentToolName = core.getToolMode().name;
+    
+      if (currentToolName=== window.Tools.ToolNames.SIGNATURE && openElements['signatureModal']) {
```
ZhijieZhang, line 470: Sorry just nit picking here, can we have a space before `===`

src/helpers/hotkeysManager.js
```diff
@@ -460,7 +461,17 @@ WebViewer(...)
...
+        // disable changing tool when the signature overlay is opened
+        return;
```
ZhijieZhang, line 472: Also I think we can just combine these two `if`s into this. Maybe something like
```javascript
if (isFocusingElement() || isSignatureModalOpen) {
  return;
}

handler(...args);
```

src/helpers/hotkeysManager.js
awy3: Thanks for pointing this out, this has been fixed

src/helpers/hotkeysManager.js
awy3: I refactor this to only have one `if`

PR #589, opened 2020-04-28
[Feature] Keeps search overlay open by default

sisardor:


ZhijieZhang:
@sisardor Can you remind me why we want to do this? Thanks!

PR #597, opened 2020-05-06
doc: explicitly mention that disableTools uses disableElements internally, and state the difference between them

ZhijieZhang:
@achenPDFTron Hopefully this clears things a bit, let me know if you have any suggestions.


achenPDFTron:
@ZhijieZhang what is the plan for `disableElements`. are we going to remove it or change the functionality or rename it?

I remember we had a conversation yesterday, how "disable" is sort of like the HTML disable where we can still see it, but we can't interact with it.

ZhijieZhang:
@achenPDFTron I don't think we will remove this API as this one is mostly used by customers. I think the plan is that if we can find a better name then we can use that one instead of `disableElements`. However I didn't receive any suggestions :(

PR #619, opened 2020-05-20
[Bugfix] Fix header overlay being offscreen

awy3:
This is for this bug

https://trello.com/c/YSdveJrl/1021-ui-releated-bug-signature-overlay-is-off

We prevent the header from having a left offset that is negative so it won't go offscreen

### DIFF
--- a/src/helpers/getOverlayPositionBasedOn.js
+++ b/src/helpers/getOverlayPositionBasedOn.js
@@ -39,5 +39,5 @@ export default (element, overlay, align = 'left') => {
     right = 'auto';
     left = buttonLeft - (overlayWidth - buttonWidth);
   }
-  return { left, right };
+  return { left: Math.max(left, 0), right };
 };
--- a/src/helpers/getOverlayPositionBasedOn.js
+++ b/src/helpers/getOverlayPositionBasedOn.js
@@ -39,5 +39,10 @@ export default (element, overlay, align = 'left') => {
     right = 'auto';
     left = buttonLeft - (overlayWidth - buttonWidth);
   }
-  return { left: Math.max(left, 0), right };
+
+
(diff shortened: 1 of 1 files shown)


src/helpers/getOverlayPositionBasedOn.js
```diff
@@ -39,5 +39,5 @@ export default (element, overlay, align = 'left') => {
     right = 'auto';
     left = buttonLeft - (overlayWidth - buttonWidth);
   }
-  return { left, right };
+  return { left: Math.max(left, 0), right };
```
ZhijieZhang, line 42: Do you need to do the same thing for `right` to make sure it won't be cut off on the right side of the bounding box?

awy3: I tested this more and found a bug that I fixed

As for `right`, I don't think we need anything. The `right` get set to "auto" or "16" so it doesn't seem to overflow. It's mainly just `left`that get set to a negative number sometime and appear off screen

src/helpers/getOverlayPositionBasedOn.js
```diff
@@ -39,5 +39,10 @@ export default (element, overlay, align = 'left') => {
     right = 'auto';
     left = buttonLeft - (overlayWidth - buttonWidth);
   }
-  return { left, right };
+
+
+  return { 
+    left: !isNaN(left) ? Math.max(left, 0) : left,
```
awy3, line 45: left can be set to "auto"

src/helpers/getOverlayPositionBasedOn.js
ZhijieZhang: I see, thanks for the clarification!

PR #622, opened 2020-05-21
[Bugfix] Programmatically updating layers won't toggle UI checkboxes

sisardor:


src/components/LayersPanel/LayersPanel.js
```diff
@@ -27,6 +27,23 @@ class LayersPanel extends React.PureComponent {
...
+  componentWillUnmount() {
+    core.removeEventListener('layersUpdated', this.onLayersUpdated.bind(this));
+  }
+
+  onLayersUpdated() {
+    const doc = core.getDocument();
+    doc.getLayersArray().then(layers => {
+      if (layers.length) {
+        this.props.setLayers(layers);
+        this.forceUpdate();
```
ZhijieZhang, line 43: Can you explain why we need a `forceUpdate` here?

src/components/LayersPanel/LayersPanel.js
```diff
@@ -27,6 +27,23 @@ class LayersPanel extends React.PureComponent {
...
   }
+  componentDidMount() {
+    core.addEventListener('layersUpdated', this.onLayersUpdated.bind(this));
+  }
+
+  componentWillUnmount() {
+    core.removeEventListener('layersUpdated', this.onLayersUpdated.bind(this));
+  }
+
+  onLayersUpdated() {
```
ZhijieZhang, line 38: Can we follow the convention in other components to use class properties?

```javascript
onLayersUpdated = () => {...}
```

And then we don't need to `.bind(this)`

src/components/LayersPanel/LayersPanel.js
```diff
@@ -27,6 +27,23 @@ class LayersPanel extends React.PureComponent {
       window.docViewer.updateView();
     }
   }
+  componentDidMount() {
+    core.addEventListener('layersUpdated', this.onLayersUpdated.bind(this));
+  }
+
+  componentWillUnmount() {
+    core.removeEventListener('layersUpdated', this.onLayersUpdated.bind(this));
```
ZhijieZhang, line 35: It looks like this depends on a WebViewer PR, so let me know if  this is ready to merge.

ZhijieZhang:
@sisardor do we still want this PR to be merged?

src/components/LayersPanel/LayersPanel.js
sisardor: Because change is coming from backend, UI is not updating, so we need to force update

ZhijieZhang: hmm ok, so does `this.props.setLayers(layers);` actually do anything? Maybe we can add a comment explaining why `this.forceUpdate` is needed here.

sisardor: setLayers is updating parent component when layers change we supposed to update parent like in line 69.

src/components/LayersPanel/LayersPanel.js
```diff
@@ -27,6 +27,25 @@ class LayersPanel extends React.PureComponent {
       window.docViewer.updateView();
     }
   }
+  componentDidMount() {
+    core.getDocument().on('layersUpdated', this.onLayersUpdated);
```
mparizeau-pdftron, line 31: It doesn't seem like this works as expected. componentDidMount is fired when the I actually open the panel so if there are any programmatic layer changes before the panel is opened then they won't be updated here.

Maybe we should be listening for this event somewhere else globally and calling the action there. Then in this component we'll just use the selector which also means we shouldn't need to use a forceUpdate?

ZhijieZhang:

Issue #610, opened 2020-05-15
Measurement Tools Bugs


elephantcatdog:
Version: 6.3.0

In the [measurement tools demo](https://www.pdftron.com/webviewer/demo/measurement-tools), an area measurement markup placed in the approximate middle of the page will have a style popup that gets cut off at the bottom. My window height is 937px.
![image](https://user-images.githubusercontent.com/37936169/82098584-d8052f00-96ca-11ea-98c9-c551f2e25ec7.png)

If the window is narrow enough that the measurement overlay overlaps the pdf, any measurement tool will be hidden behind the measurement overlay while drawing. The area and perimeter tools will not allow a node to be placed behind the overlay at all.
![image](https://user-images.githubusercontent.com/37936169/82100263-e7867700-96ce-11ea-9b8b-a6999447139e.png)
Consider allowing the overlay to become transparent or move when the user is drawing and hovers on it.

I will add any other measurement issues as I find them. Thank you.

ZhijieZhang:
Thanks for the feedback!

For the first issue I think we will resolve it by just making the popup draggable. So in this case you can drag it to somewhere else to see the `Precision`.

The second issue might take some time to resolve. I will need to poke around.

henry-chris:
@ZhijieZhang  As a quick addendum. 

In the demo you have a 'Enable Snapping' option.  We are placing that option directly in the tool popup as a check box (but not in the annotationStylePopup).  I think in the 'MeasurementOptions'.

Do you think that's a valuable option to put in stock webviewer?  I ask because we like to have as few changes as possible to the standard viewer.

elephantcatdog:
Hi @ZhijieZhang,
I've found another issue.

When snapping is enabled on a tool, the annotation snaps to lines during the initial drawing, but not when editing it. This occurs for any tool with snapping enabled, as in the [snapping demo](https://www.pdftron.com/webviewer/demo/snap-to-objects), not just measurement tools.

I also noticed that when using the area or perimeter tools (or any tools that use clicks to place a point), both left-clicking and double-clicking can end an annotation. Using left-clicking to end the annotation also brings up the left-click context menu. It would be nice if this menu didn't appear when ending an annotation, as it could be confusing to users when they expect the annotation menu.
![image](https://user-images.githubusercontent.com/37936169/82500396-ec756d00-9ab8-11ea-9224-9f3ca8499150.png)
[...]

PR #628, opened 2020-05-26
feat(MeasurementOverlay): allow creation and editing measurement annotations underneath the measurement overlay

ZhijieZhang:
The purpose of this PR is to address the second feedback from https://github.com/PDFTron/webviewer-ui/issues/610.

This PR does two things:
1. Make the measurement overlay component draggable, so users will be able to move it around in the view to find a comfortable spot.
2. Make the component transparent when
    - we are creating an annotation and our mouse is hovering the component
    - we are selecting an annotation and our mouse is hovering on the part of the annotation that's under the component
Before this PR, a user won't be able to create a path point under the component for polyline and polygon annotations, which I see can be a bit annoying.

Also some of the code is formatted.

@bollain Could you take a look when you have time? Just want to see if you have any feedback.

### DIFF
--- a/src/components/MeasurementOverlay/MeasurementOverlay.js
+++ b/src/components/MeasurementOverlay/MeasurementOverlay.js
@@ -1,4 +1,6 @@
 import React from 'react';
+import classNames from 'classnames';
+import Draggable from 'react-draggable';
 import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
 import { connect } from 'react-redux';
@@ -6,12 +8,7 @@ import { connect } from 'react-redux';
 import Icon from 'components/Icon';
 
 import core from 'core';
-import {
-  mapAnnotationToKey,
-  mapToolNameToKey,
-  getDataWithKey,
-} from 'constants/map';
-import getClassName from 'helpers/getClassName';
+import { mapAnnotationToKey, mapToolNameToKey, getDataWithKey } from 'constants/map';
 import parseMeasurementContents from 'helpers/parseMeasurementContents';
 import actions from 'actions';
 import selectors from 'selectors';
@@ -36,7 +33,14 @@ class MeasurementOverlay extends React.PureComponent {
     super(props);
     this.state = {
       annotation: null,
+      transparentBackground: false,
+      position: {
+        x: 0,
+        y: 0,
+      },
     };
+    this.overlayRef = React.createRef();
+    this.isCreatingAnnotation = false;
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


src/components/MeasurementOverlay/MeasurementOverlay.js
```diff
@@ -36,7 +33,14 @@ class MeasurementOverlay extends React.PureComponent {
     super(props);
     this.state = {
       annotation: null,
+      transparentBackground: false,
```
ZhijieZhang, line 36: @bollain Let me know if the usage of this state and `this.isCreatingAnnotation` is clear and easy to understand. If not I can write some comments in the code to explain.

src/components/MeasurementOverlay/MeasurementOverlay.js
```diff
@@ -299,46 +345,63 @@ class MeasurementOverlay extends React.PureComponent {
...
+            MeasurementOverlay: true,
+            open: isOpen,
+            closed: !isOpen,
+            transparent: transparentBackground,
+          })}
+          ref={this.overlayRef}
+          data-element="measurementOverlay"
+        >
+          {this.shouldShowCustomOverlay(annotation) ? (
+            <CustomMeasurementOverlay
```
ZhijieZhang, line 375: I decided to move this component and `EllipseMeasurementOverlay` into the rendered component so that I don't need to import and use Draggable in three places

src/components/MeasurementOverlay/MeasurementOverlay.js
```diff
@@ -66,21 +74,49 @@ class MeasurementOverlay extends React.PureComponent {
...
-  onMouseMove = () => {
+  onMouseMove = e => {
     const { activeToolName, openElement } = this.props;
     const tool = core.getTool(activeToolName);
 
     if (this.state.annotation) {
+      const insideRect = this.isMouseInsideRect(e, this.overlayRef.current);
+      let useTransparentBackground;
+
+      if (this.isCreatingAnnotation) {
```
ZhijieZhang, line 85: This is the main logic to control if we should render the component transparent. Let me know if you have any questions.

src/components/MeasurementOverlay/MeasurementOverlay.js
bollain: I havent discussed this with Matt recently, but we may just get rid of the CustomMeasurementOverlay since we ended up implementing the EllipseMeasurement on our own.

bollain: Makes sense to move this here, it reduces readability but I think if we remove the CustomMeasurementOverlay this will look better.

ZhijieZhang: I see. Yeah I think we can actually remove the custom overlay. I will do that in another PR.

ZhijieZhang:
@bollain Thanks for reviewing. I'm going to merge this in, and remove the custom overlay in another commit.

PR #636, opened 2020-05-28
[Express] Set up window params for express

lbittner-pdftron:
Part of https://github.com/XodoDocs/webviewer/pull/1272


lbittner-pdftron:
NVM going to do it different

PR #640, opened 2020-06-02
[Feature] Print current view option and null checks

awy3:
Added an option for printing the current view. Need a WebViewer update for this to work because the "renderRect" option doesn't currently work

image

### DIFF
--- a/src/components/PrintModal/PrintModal.js
+++ b/src/components/PrintModal/PrintModal.js
@@ -62,7 +62,6 @@ class PrintModal extends React.PureComponent {
     if(!selectedPrintOption?.current){
       const printOptions = [this.allPages, this.currentPage, this.currentView, this.customPages];
       const selectedOptions = printOptions.find(o => o.current?.checked)
-      
       this.setState({ selectedPrintOption: selectedOptions ?? printOptions.find(o => o?.current)});
     }
     
@@ -163,7 +162,7 @@ class PrintModal extends React.PureComponent {
 
     const creatingPages = this.creatingPages();
     Promise.all(creatingPages)
-      .then(pages => {        
+      .then(pages => {
         this.printPages(pages);
         this.resetPrintQuality();
       })
@@ -239,7 +238,7 @@ class PrintModal extends React.PureComponent {
         const x2 = Math.max(coordinates[0].x, coordinates[1].x);
         const y2 = Math.max(coordinates[0].y, coordinates[1].y);
 
-        zoom = this.props.zoom;
+        zoom = core.getZoom();
         renderRect = { x1, y1, x2, y2 };
       }
 
@@ -248,7 +247,7 @@ class PrintModal extends React.PureComponent {
         zoom,
         pageRotation: printRotation,
         drawComplete: onCanvasLoaded,
-        'renderRect': renderRect,
+        renderRect,
       });
       this.pendingCanvases.push(id);
     });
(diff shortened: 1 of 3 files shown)


ZhijieZhang:
Can you open PRs to `master` and `legacy/master` as well? Just want to make sure that this feature won't be missing in the newer version.

src/components/PrintModal/PrintModal.js
```diff
@@ -52,10 +53,18 @@ class PrintModal extends React.PureComponent {
       isWatermarkModalVisible: false,
       watermarkModalOption: null,
       existingWatermarks: null,
+      selectedPrintOption: this.allPages,
```
ZhijieZhang, line 56: I think my question regarding this PR is about this change. I'm actually not sure if it's a common or good practice to store a React ref on a React state. Can you explain a little bit more about why we are doing this? Specifically, what problem did you find in the previous implementation, and how does the implementation here resolve that problem?

The `Input` component can be a controlled or uncontrolled component depending on the prop usage. If I'm understanding correctly, in this PR we are trying to make the `Input`a controlled component. That is, the PrintModal component has the full control of whether a print option input is checked or not. The `Input` component in the previous implementation is uncontrolled, and we just listen to the onChange event. I guess I'm not clear about the benefits of making them be controlled?

src/components/PrintModal/PrintModal.js
```diff
@@ -610,6 +663,10 @@ const mapStateToProps = state => ({
   sortStrategy: selectors.getSortStrategy(state),
   colorMap: selectors.getColorMap(state),
   layoutMode: selectors.getDisplayMode(state),
+  printAllPageDisabled: selectors.isElementDisabled(state, 'allPagesPrintOption'),
+  printCurrentPageDisabled: selectors.isElementDisabled(state, 'currentPagePrintOption'),
+  printViewPageDisabled: selectors.isElementDisabled(state, 'currentViewPrintOption'),
+  printCustomPageDisabled: selectors.isElementDisabled(state, 'customPagesPrintOption'),
```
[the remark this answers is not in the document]
awy3: this is so if the user does "disableElements", the default option won't point to a option that is hidden

src/components/PrintModal/PrintModal.js
```diff
@@ -610,6 +658,7 @@ const mapStateToProps = state => ({
   sortStrategy: selectors.getSortStrategy(state),
   colorMap: selectors.getColorMap(state),
   layoutMode: selectors.getDisplayMode(state),
+  disabledPrintOptionCount: selectors.getDisabledElementsCountFromList(state, ['allPagesPrintOption', 'currentPagePrintOption', 'currentViewPrintOption', 'customPagesPrintOption']),
```
ZhijieZhang, line 661: Why do we need this?

src/components/PrintModal/PrintModal.js
```diff
@@ -56,6 +58,17 @@ class PrintModal extends React.PureComponent {
   }
 
   componentDidUpdate(prevProps) {
+    const printOptions = [this.allPages, this.currentPage, this.currentView, this.customPages];
+    const selectedPrintOption = printOptions.find(printOption => printOption.current?.checked);
+
+    if(!selectedPrintOption){
```
ZhijieZhang, line 64: Can you explain why we need this?

awy3:
Customer no longer needed this

PR #643, opened 2020-06-02
Basic a11y

liamross:
First pass at making buttons show tab indicators, and incorporating the toolkit in for checkboxes (right now only in search overlay).


liamross:
We should probably squash this when we merge it, the only reason so many commits are showing up is cause @kristianhein 's new ui branch was probably squashed into master, and this is technically branched off of that branch prior to it merging with master

src/components/Input/Input.js
```diff
@@ -1,12 +1,11 @@
 import React from 'react';
-import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import Icon from 'components/Icon';
+import { useSelector } from 'react-redux';
```
ZhijieZhang, line 4: You don't like where it was? Just curious, no need to revert the change.

src/components/SearchOverlay/SearchOverlay.js
```diff
@@ -459,24 +462,6 @@ class SearchOverlay extends React.PureComponent {
...
-    this.executeDebouncedFullSearch();
-  }
-
-  onChangeWildcard = e => {
-    this.props.setWildcard(e.target.checked);
-    this.clearSearchResults();
-    this.executeDebouncedFullSearch();
-  }
-
-  onChangeWildcard = e => {
```
ZhijieZhang, line 474: Good catch!

src/components/SearchOverlay/SearchOverlay.js
```diff
@@ -5,8 +5,6 @@ import { connect } from 'react-redux';
 import onClickOutside from 'react-onclickoutside';
 
 import Icon from 'components/Icon';
-import Tooltip from 'components/Tooltip';
-import Input from 'components/Input';
```
ZhijieZhang, line 9: Do we want to use the toolkit for the `PrintModal` as well? That component uses `Input` as well.

ZhijieZhang:
Nothing is stopping this PR from being merged. Just have few questions.

src/components/SearchOverlay/SearchOverlay.js
liamross: Yeah I'll do that one next, just slightly more complexity with it lining up properly

src/components/Input/Input.js
liamross: haha this one was an accident, I sometimes instinctively hit <kbd>Shift</kbd> + <kbd>⌘</kbd> + <kbd>O</kbd> which organizes the imports, let me revert

ZhijieZhang:
LGTM, thanks!

PR #650, opened 2020-06-03
Add API to customize annotation tooltips

bollain:
Trello card:
https://trello.com/c/OsViPPT3/906-custom-tooltips

Adds new API so users can override current tooltip behaviour on annotations.

API takes a callback that should return a DOM element.
```
Webviewer({
  initialDoc: hashFile,
  path: '/lib',
}, document.getElementById('viewer')).then(instance => {
  instance.setAnnotationContentOverlayHandler(annotation => {
    var div = document.createElement("div")
    div.appendChild(document.createTextNode(`Created by: ${annotation.Author}`))
    div.appendChild(document.createElement('br'))
    div.appendChild(document.createTextNode(`Created on ${annotation.DateCreated}`))
    return div
  })
});
```

### DIFF
--- a/src/apis/setAnnotationContentOverlayHandler.js
+++ b/src/apis/setAnnotationContentOverlayHandler.js
@@ -1,5 +1,22 @@
 import action from 'actions';
 
+/**
+ * Adds a custom overlay to annotations if that annotation currently support it. Currently only the Ellipsis annotation supports it.
+ * @method WebViewerInstance#setAnnotationContentOverlayHandler
+ * @param {function} customOverlayHandler a function that takes an annotation and returns a DOM Element, which is rendered as a tooltip when hovering over the annotation
+ *  * @example
+WebViewer(...)
+  .then(function(instance) {
+    instance.setAnnotationContentOverlayHandler(annotation => {
+        var div = document.createElement("div")
+        div.appendChild(document.createTextNode(`Created by: ${annotation.Author}`))
+        div.appendChild(document.createElement('br'))
+        div.appendChild(document.createTextNode(`Created on ${annotation.DateCreated}`))
+        return div
+    })
+  });
+ */
+
 export default store => annotationContentOverlayHandler => {
   store.dispatch(action.setAnnotationContentOverlayHandler(annotationContentOverlayHandler));
 };
--- a/src/apis/setAnnotationContentOverlayHandler.js
+++ b/src/apis/setAnnotationContentOverlayHandler.js
@@ -1,19 +1,19 @@
 import action from 'actions';
 
 /**
- * Adds a custom overlay to annotations if that annotation currently support it. Currently only the Ellipsis annotation supports it.
+ * Adds a custom overlay to annotations on mouseHover, overriding the existing overlay.
  * @method WebViewerInstance#setAnnotationContentOverlayHandler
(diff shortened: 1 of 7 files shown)


src/apis/setAnnotationContentOverlayHandler.js
```diff
@@ -0,0 +1,22 @@
+import action from 'actions';
+
+/**
+ * Adds a custom overlay to annotations if that annotation currently support it. Currently only the Ellipsis annotation supports it.
```
ZhijieZhang, line 4: Hmm maybe we forgot to change the description? ( •̀ .̫ •́ )✧

src/apis/setAnnotationContentOverlayHandler.js
```diff
@@ -0,0 +1,22 @@
...
+ * @method WebViewerInstance#setAnnotationContentOverlayHandler
+ * @param {function} customOverlayHandler a function that takes an annotation and returns a DOM Element, which is rendered as a tooltip when hovering over the annotation
+ *  * @example
+WebViewer(...)
+  .then(function(instance) {
+    instance.setAnnotationContentOverlayHandler(annotation => {
+        var div = document.createElement("div")
```
ZhijieZhang, line 11: Let's make the code snippets a bit more consistent. Can we use `const`, `let`, single quote for strings, and end the function call with a `;`?

src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
```diff
@@ -55,6 +64,20 @@ const AnnotationContentOverlay = () => {
   const contents = annotation?.getContents();
   const numberOfReplies = annotation?.getReplies().length;
 
+  if (annotation && isUsingCustomHandler) {
```
ZhijieZhang, line 67: Hmm I think we still want to have the `isDisabled` and `isMobileDevice` check for the custom tooltip. What do you think?

Also I think the code would look a bit more better if we can put the `CustomElement` inside of the `<div>` below. For example:
```javascript
// may need to pass contents to <DefaultOverlay>, or maybe just pass `annotation`
return isDisabled || isMobileDevice || !contents ? null : (
  <div
    className="Overlay AnnotationContentOverlay"
    data-element="annotationContentOverlay"
    style={{ ...overlayPosition }}
  >
    {isUsingCustomHandler ? <CustomElement /> : <DefaultOverlay ... />}
  </div>
);

const DefaultOverlay = (props) => (
  <>
    <div className="author">{core.getDisplayAuthor(annotation)}</div>
    <div className="contents">
      {contents.length > MAX_CHARACTERS
        ? `${contents.slice(0, MAX_CHARACTERS)}...`
        : contents}
    </div>
    {numberOfReplies > 0 && (
      <div className="replies">
        {t('message.annotationReplyCount', { count: numberOfReplies })}
      </div>
    )}
  </>
)
```

src/apis/setAnnotationContentOverlayHandler.js
bollain: My bad!

src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
bollain: @ZhijieZhang Refactored things a bit, little different from your proposal due to the `contents` part of the conditional rendering, but I think the code is a bit cleaner.

PR #653, opened 2020-06-05
a11y updates

liamross:


src/components/Note/Note.js
```diff
@@ -104,6 +104,14 @@ const Note = ({ annotation }) => {
 
   const showReplyArea = !Object.values(isEditingMap).some(val => val);
 
+  const handleNoteKeydown = e => {
+    // Stop enter/space key on Note from being submitted into field.
+    const isNote = e.target === e.currentTarget;
+    if (isNote && (e.key === 'Enter' || e.key === ' ')) {
+      e.preventDefault();
+    }
```
liamross, line 112: Checks if target issuing key events is the same as Note (`currentTarget`), if it is prevent <kbd>return</kbd> and <kbd>space</kbd> from doing default behaviour since upon opening the Note, this would be entered into the input field for commenting

src/components/Tooltip/Tooltip.js
```diff
@@ -13,9 +13,10 @@ const propTypes = {
   hideShortcut: PropTypes.bool,
 };
 
-const Tooltip = forwardRef( ({ content = '', children, hideShortcut }, forwardedRef) => {
+const Tooltip = forwardRef( ({ content = '', children, hideShortcut }, ref) => {
   const timeoutRef = useRef(null);
-  const childRef = forwardedRef ? forwardedRef : useRef(null);
+  const childRef = useRef(null);
+  useImperativeHandle(ref, () => childRef.current);
```
liamross, line 19: Fixes conditional use of `useRef` (shouldn't conditionally use hooks, can mess with the flow although in this case it's dependent on ref so probably safeish). Also, this is the proper way to use a ref internally but also have it output with `forwardRef` (unfortunately also not very well documented).

ZhijieZhang: Ah this is pretty cool. Thanks for showing this.

PR #655, opened 2020-06-08
[6.3][Bugfix] Fixed sliders to use data-element

S0ulDrag0n:
Sliders were provided data-element values but were not actually used. This PR should correct that, as well as a bug where disabling one, would cause the next to render due to a mismatch with the slider prop indices.

I could not wrap the sliders in another element to add this data-element since the elements would then not line up in the grid.

image


S0ulDrag0n:
Yea, I will prepare one tomorrow.

S0ulDrag0n:
#659

S0ulDrag0n:
#660

PR #656, opened 2020-06-09
[Feature][Legacy UI] Custom stamp tab for Rubber stamp overlay

sisardor:


src/components/RubberStampToolButton/RubberStampToolButton.js
```diff
@@ -0,0 +1,40 @@
...
+import { useDispatch, useSelector, shallowEqual } from 'react-redux';
+
+import Button from 'components/Button';
+
+import core from 'core';
+import actions from 'actions';
+import selectors from 'selectors';
+
+const RubberStampToolButton = () => {
+  const [isSignatureOverlayOpen] = useSelector(
```
ZhijieZhang, line 12: Can we rename this variable name?

src/components/RubberStampToolButton/RubberStampToolButton.js
```diff
@@ -0,0 +1,40 @@
...
+    'down-arrow': true,
+  });
+
+  return (
+    <Button
+      className={buttonClass}
+      isActive={isSignatureOverlayOpen}
+      img="ic_annotation_stamp_black_24px"
+      onClick={handleClick}
+      title="annotation.rubberStamp"
```
ZhijieZhang, line 35: I noticed that the tooltip of this button doesn't contain a hotkey. I think we should add a hotkey for this button to match other buttons in the header. Do you think this makes sense?

src/components/StampOverlay/StampOverlay.js
```diff
@@ -50,14 +53,17 @@ class StampOverlay extends React.Component {
...
+        'menuOverlay',
+        'toolsOverlay',
+        'zoomOverlay',
+        'toolStylePopup',
+      ]);
       this.setOverlayPosition();
       this.getDefaultRubberStamps(isLanChanged);
-    } else if (this.props.isOpen && isLanChanged) {
-      this.getDefaultRubberStamps(isLanChanged);
+      this.getCustomRubberStamps(isLanChanged);
```
ZhijieZhang, line 66: I feel like we don't need to call `getDefaultRubberStamps` and `getCustomRubberStamps` every time the overlay opens. Maybe just when the language changes?

Do you think it makes sense to have these two function calls in `componentDidMount` and then in `componentDidUpdate` we only call them when the language changes? I think this can improve some performance, and is a relative change that we can do.

src/components/StampOverlay/StampOverlay.js
```diff
@@ -108,6 +114,32 @@ class StampOverlay extends React.Component {
     this.setState({ isStampSelected: true });
   }
 
+  getCustomRubberStamps = async isLanChanged => {
+    if (!this.state.customAnnotations.length || isLanChanged) {
+      const annotations = await this.stampTool.getCustomStampAnnotations();
+      const previews = await Promise.all(
```
ZhijieZhang, line 120: Do we need this `previews`? I don't see it being used in this function.

src/components/StampOverlay/StampOverlay.js
```diff
@@ -166,11 +211,33 @@ class StampOverlay extends React.Component {
...
+                <Button label={StandardBusiness} />
+              </Tab>
+              <Tab dataElement="customRubberStampButton">
+                <Button label={CustomStamp} />
+              </Tab>
+            </div>
           </div>
-        </div>
+
+          <TabPanel dataElement="defaultRubberStamp">
```
ZhijieZhang, line 226: Can we follow the naming conventions for other `TabPanel`s and end the `dataElement` with `Panel`? For example, we can have `defaultRubberStampPanel` here.

src/components/StampOverlay/StampOverlay.js
```diff
@@ -166,11 +211,33 @@ class StampOverlay extends React.Component {
...
+            <div className="default-stamp-container">
+              <div className="modal-body">
+                { imgs }
+              </div>
+            </div>
+          </TabPanel>
+          <TabPanel dataElement="customRubberStamp">
+            <div className="custom-stamp-container">
+              <div className="modal-body">
```
ZhijieZhang, line 235: DO we need this `modal-body` div? I don't think we have any css styles for it.

src/components/StampOverlay/StampOverlay.js
```diff
@@ -108,6 +114,32 @@ class StampOverlay extends React.Component {
...
+          return this.stampTool.getPreview(annotation, options);
+        }),
+      );
+
+      const customAnnotations = annotations.map(annotation => ({
+        annotation,
+        imgSrc: annotation['ImageData'],
+      }));
+
+      this.setState({ customAnnotations });
```
ZhijieZhang, line 139: In the `getDefaultRubberStamps` function we have `this.setState({ defaultAnnotations, language: this.props.i18n.language });`, can you explain a little bit here about why we don't need set the state for `language` in this function?

Actually do you need to store `language` on the state? I don't see it being used in the render function.

i18n/translation-en.json
```diff
@@ -307,7 +307,9 @@
     "pan": "Pan",
     "select": "Select",
     "Marquee": "Marquee Zoom",
-    "Link": "Link URL or Page"
+    "Link": "Link URL or Page",
+    "StandardBusiness": "Standard Business",
```
ZhijieZhang, line 311: When you have time can you follow the wiki https://github.com/XodoDocs/webviewer/wiki/WebViewer-UI-Localization to translate them into other languages? Thanks.

src/components/StampOverlay/StampOverlay.js
sisardor: Yes I think that makes sense. My original code was doing that, and later it got changed to something else. I will check language change and update accordingly

src/components/StampOverlay/StampOverlay.js
sisardor: Here text for a dynamic stamp is entered by the user, and the user can enter it in any language, so we don't need to translate.

The this.state.language is used in componentDidUpdate in order to check if language changed at any time.

ZhijieZhang:
@sisardor What approach did you use to update the translation files? They look weird to me. The script shouldn't update the translations for existing keys.

ZhijieZhang:
@sisardor Can you let me know which WebViewer branch I should use to test this feature out? Thanks.

sisardor:
> Thanks for the updates! I think the code looks good to me!
> 
> However as I mentioned above the translation files don't look correct. We need to fix it until this PR can be merged in.

I added new entries to English json and I just used `npm run translation`, I'll double check and fix it.

ZhijieZhang:
> > Thanks for the updates! I think the code looks good to me!
> > However as I mentioned above the translation files don't look correct. We need to fix it until this PR can be merged in.
> 
> I added new entries to English json and I just used `npm run translation`, I'll double check and fix it.

It seems good now. Feel free to merge it yourself when the WebViewer PR is in.

PR #661, opened 2020-06-11
[CHANGE] fixed issue where it scrolled to incorrect row when clicking…

achenPDFTron:
fixes https://support.pdftron.com/a/tickets/14300

### DIFF
--- a/src/components/ThumbnailsPanel/ThumbnailsPanel.js
+++ b/src/components/ThumbnailsPanel/ThumbnailsPanel.js
@@ -262,18 +262,9 @@ class ThumbnailsPanel extends React.PureComponent {
   };
 
   onPageNumberUpdated = pageNumber => {
-    const {
-      width,
-    } = this.state;
-    const numberOfColumns = this.getNumberOfColumns(width);
-    // from doing some drawings, page numbers are always ordered from left to right in asc order
-    new Array(numberOfColumns).fill().forEach((columnIndex) => {
-      if (pageNumber % (columnIndex + 1) === 0) {
-        // derived from this formula thumbIndex = index * numberOfColumns + columnIndex;
-        const rowIndex = ((pageNumber - 1) - columnIndex) / numberOfColumns;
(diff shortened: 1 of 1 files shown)


src/components/ThumbnailsPanel/ThumbnailsPanel.js
```diff
@@ -262,7 +262,18 @@ class ThumbnailsPanel extends React.PureComponent {
...
 
   onPageNumberUpdated = pageNumber => {
-    this.listRef.current?.scrollToRow(pageNumber - 1);
+    const {
+      width,
+    } = this.state;
+    const numberOfColumns = this.getNumberOfColumns(width);
+    // from doing some drawings, page numbers are always ordered from left to right in asc order
+    new Array(numberOfColumns).fill().forEach((columnIndex) => {
+      if (pageNumber % (columnIndex + 1) === 0) {
```
ZhijieZhang, line 271: It seems that this `if` check will always be false because `columnIndex` is always `undefined`.
Can you explain a little bit more about `from doing some drawings?`, I don't think I see any `drawings` in the comments. Maybe I'm misunderstanding something?

Also I feel like the logic can be simplied to
```js
const numberOfColumns = this.getNumberOfColumns(this.state.width);
const pageIndex = pageNumber - 1;
this.listRef.current?.scrollToRow(Math.floor(pageIndex / numberOfColumns));
```

achenPDFTron: ```Can you explain a little bit more about from doing some drawings?```
I drew a few examples in my notebook before I came up with a formula.

But i think your formula is way easier to read compared to mines. so yes, i think we can use your suggestion.

achenPDFTron: i have removed the comment also as it might cause some confusions if you got confused.

achenPDFTron:
@ZhijieZhang  done https://github.com/PDFTron/webviewer-ui/pull/666
https://github.com/PDFTron/webviewer-ui/pull/667

PR #665, opened 2020-06-12
Tooltip perf

bollain:
Made some changes based on client feedback. CustomOverlay was re-rendering on every mouseMove, which made jitter as there was an expensive DOM manipulation operation happening everytime.

To fix this, only render the custom overlay once, by making sure the references stay the same.

Also adds a check so CustomOverlays show for text annotations, as the client requests this.

To test:
```
  instance.setAnnotationContentOverlayHandler(annotation => {
    const div = document.createElement("div")
    div.appendChild(document.createTextNode(`Created by: ${annotation.Author}`))
    div.appendChild(document.createElement('br'))
    div.appendChild(document.createTextNode(`Created on ${annotation.DateCreated}`))
    return div
  })
```

### DIFF
--- a/src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
+++ b/src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
@@ -1,5 +1,5 @@
 /* eslint-disable react/prop-types */
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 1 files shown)


src/components/AnnotationContentOverlay/AnnotationContentOverlay.js
```diff
@@ -109,10 +101,18 @@ const AnnotationContentOverlay = () => {
     }
   };
 
-  if (isDisabled || isMobileDevice) {
+  if (isDisabled || isMobileDevice || !annotation) {
     return null;
   } else if (isUsingCustomHandler) {
-    return <CustomOverlay />;
+    return (
+      <div
```
ZhijieZhang, line 108: Do you think we should do the same for `DefaultOverlay`? I didn't actually test it but I think we also keep unmounting and then mounting them back because it's a new component every time we re-render. Since the DefaultOverlay is what most of the clients are using I think it's worth taking a further look.

bollain: It is indeed mounting and unmounting the `DefaultOverlay`on every mouse move;  code looks not as clean but this should take care of it.

PR #669, opened 2020-06-12
[Feature] New Tab for dynamic custom stamp

sisardor:


ZhijieZhang:
Hi Sardor, It seems that this PR is merging into another new feature and I have a little bit of trouble figuring out which branch in the WebViewer repo I should check out so that I can see this feature working. Can you let me know the branch name? Thanks.

PR #671, opened 2020-06-15
feat(a11y): refactor menu overlay to prepare for a11y

liamross:
- converted to functional component (reduced lines from 255 --> 143)

I did this cause it simplifies the component in order to allow for easier dropdown-style menu a11y. Additionally, it will be easier to extract that logic and move into other components if we need to make more dropdown-style menus accessible.

I feel like we should be pushing for this pattern more:
- No PropTypes which is nicer
- Everything is self contained in the component
- Much smaller code-wise (reduction of over 100 lines)
- Easier TypeScript conversion if we choose to go that route in the future


liamross:
@ZhijieZhang

> For functional components I feel like we usually just do a destructure at the beginning of the component, so without PropTypes it's still clear to use what the props the component takes in.

The issue is now there are no props, they are being created within the component (through the use of Redux hooks)

> Can you explain a little bit about the reason for using `useCallback` around `dispatch` to create those functions?

Honestly probably not necessary, more just to keep the code roughly the same (instead of function props now they are bound to the same name variables using `useCallback`. They could be removed easily (might save it for my next PR though).

PR #672, opened 2020-06-15
[Feature] count measurement annotation

JimmyLin39:
Added a count measurement tool under the measurement tools. It is a special type of annotation that when selected it will show many instances of it exist in the document. It's using the sticky create tool behind the scenes


src/components/MeasurementOverlay/CustomMeasurementOverlay.js
```diff
@@ -9,6 +9,8 @@ function CustomMeasurementOverlay(props) {
   const renderAppropriateOverlay = type => {
     const annotationType = mapAnnotationToKey(props.annotation);
     switch (annotationType) {
+      case 'countMeasurement':
+        return <CustomCountMeasurementOverlay {...props}/>;
```
ZhijieZhang, line 13: Ah sorry I forgot to mention this component. The `CustomMeasurementOverlay` was implemented for a particular customer who would like to see some different contents in the component when a ellipse measurement annotation is selected. Eventually we found that we will just make it built-in in the UI, which is the `EllipseMeasurementOverlay` component you can find in `MeasurementOverlay.js`. That means, the `CustomMeasurementOverlay` is not being used by anyone and is not showing up by default now, and we will eventually delete this componnet.

Since we are making the `countMeasurement` built-in in WebViewer, do you think it makes more sense to move this overlay to `MeasurementOverlay`? You can take a look at how it renders `EllipseMeasurementOverlay`.

i18n/translation-en.json
```diff
@@ -52,6 +52,7 @@
     "callout": "Callout",
     "crop": "Crop Page",
     "caret": "Caret",
+    "countMeasurement": "Count measurement",
```
ZhijieZhang, line 55: Can you follow [this wiki](https://github.com/XodoDocs/webviewer/wiki/WebViewer-UI-Localization) to generate the text for other languages?

src/components/MeasurementOverlay/CustomMeasurementOverlay.js
JimmyLin39: Sounds good, I will create a separate `CountMeasurementOverlay.js` and modify some rendering logic in the `MeasurementOverlay.js`.

PR #675, opened 2020-06-16
Bugfix/svg missing attributes

CorreyL:
Hello webviewer Maintainers,

I'm submitting this fix to ensure that the `assets/icons/ic_annotation_rectangular_area_black_24px.svg` file renders at the expected `width` and `height` of `24px`.

Here's what happens when you look at the `.svg` on `GitHub` on `Chrome`:

image

I've also edited the formatting to improve readability.

# Screenshots

### Before:

image

### After:

image

### Before:

image

### After:

image


CorreyL:
In-case we ever need to return to a similar issue, and add the `height` and `width` attributes at scale:

```python
from html.parser import HTMLParser
import glob
import re

# Sample code from https://docs.python.org/3/library/html.parser.html
class MyHTMLParser(HTMLParser):
    def __init__(self):
        super().__init__()
        self.count = 0
        self.current_filename = None

    def handle_starttag(self, tag, attrs):
        if tag == "file":
            # <file name="foobar">
            self.current_filename = attrs[0][1]
        if tag == "svg":
            attr_collector = {}
            for attr_tuple in attrs:
                name = attr_tuple[0]
                value = attr_tuple[1]
                attr_collector[name] = value
            if "height" not in attr_collector and "width" not in attr_collector:
                print(
                  f"Could not find svg height and width in {self.current_filename}"
                )

    def handle_endtag(self, tag):
        pass

    def handle_data(self, data):
        pass

if __name__ == "__main__":
    svg_files = glob.glob("*.svg")
    parser = MyHTMLParser()

    for svg_file in svg_files:
        with open(svg_file) as svg_content:
          lines = svg_content.readlines()
          for line in lines:
            # The way MyHTMLParser is written, the only way to pass in the
            # filename is as a fake attribute of the content of the svg file,
            # since MyHTMLParser is only instantiated once at the beginning of
            # this script
            # Perhaps we can instantiate MyHTMLParser on each iteration of the
            # for loop, and pass a `filename` parameter, and set it as a `self`
            # variable?
            parser.feed(f'<file name="{svg_file}">{line}')
```

Sample output as of c94a618:

```bash
$ python check-svg-attributes.py
Could not find svg height and width in icon - header - mode - day.svg
Could not find svg height and width in icon - header - mode - night.svg
Could not find svg height and width in icon - header - select - line.svg
Could not find svg height and width in icon - tool - pen and shape - phone - line.svg
# Snipped
```

PR #681, opened 2020-06-17
Feature/print option annotation

JimmyLin39:
Add an option to print without annotations. The option to print with annotation is enabled by default. 
https://trello.com/c/AsrswWPC/1028-add-option-to-print-without-annotations


i18n/translation-en.json
```diff
@@ -146,7 +146,7 @@
     "passwordRequired": "Password required",
     "preparingToPrint": "Preparing to print...",
     "annotationReplyCount": "{{count}} Reply",
-    "annotationReplyCount_plural" : "{{count}} Replies",
+    "annotationReplyCount_plural": "{{count}} Replies",
```
ZhijieZhang, line 149: Let's follow https://github.com/XodoDocs/webviewer/wiki/WebViewer-UI-Localization to have the text in other languages.

src/components/PrintModal/PrintModal.js
```diff
@@ -557,6 +559,16 @@ class PrintModal extends React.PureComponent {
                       disabled={isPrinting}
                       center
                     />
+                    <Choice
+                      dataElement="annotationsPrintOption"
+                      id="include-annotations"
+                      name="annotations"
+                      label={t('option.print.includeAnnotations')}
+                      disabled={isPrinting}
+                      onChange = {() => this.setState(state => ({ includeAnnotations: !state.includeAnnotations }))}
```
ZhijieZhang, line 568: Hmm do you think to make the code look consistent we should make it an uncontrolled component like other `<Choice>` and use `ref.current.checked` instead? I'm ok with this.

src/components/PrintModal/PrintModal.js
```diff
@@ -190,23 +191,24 @@ class PrintModal extends React.PureComponent {
...
-            this.setState(({ count }) => ({
-              count: count < 0 ? -1 : count + 1,
-            }));
-            resolve(img);
-          };
-        });
-      };
 
+        if (this.state.includeAnnotations) {
+          await this.drawAnnotationsOnCanvas(canvas, pageNumber);
```
ZhijieZhang, line 201: The `drawAnnotationsOnCanvas` function actually does two things. It draws all the markup annotations, such as highlight, comment, freehand, etc on the canvas, and it also draws all the widget annotations on it. I think when `includeAnnotations` is not checked, we don't want to see markup annotations, but still want to see widget annotations. Can you check with Sardor(the trello card creater) to see if this is the case?

If this is true, then you probably need to tweak the logic a bit here, since currently if the checkbox is unchecked, widget annotations don't appear in the print document. Let me know if you have any questions. Thanks!

src/components/PrintModal/PrintModal.js
JimmyLin39: I am not able to use the `ref.current.checked` if it needs to be `checked` by default. Maybe its because it warps in the <Choice /> component? [react-toolkit Choice component](https://pdftron.github.io/webviewer-react-toolkit/?path=/story/components-choice--controlled)

ZhijieZhang: I see. I think the rest of the props aren't documented. I checked the code and it seems that it will pass through most of the props to the <input> element.

If you look at the first `Choice` in this file it takes in a `defaultChecked` prop. I think you can try it and see if it works?

JimmyLin39: Sure, I will try it. Thanks!

JimmyLin39: The `defaultChecked` seems not working for the checkbox. @liamross

src/components/PrintModal/PrintModal.js
JimmyLin39: Sardor confirmed that
> this is option basically to print PDF without all markup annotations and widgets

src/components/PrintModal/PrintModal.js
ZhijieZhang: ok I see. Let's keep it this way then.

src/components/PrintModal/PrintModal.js
ZhijieZhang: Thanks for the info. I talked to Sardor and read the discussion in the ticket. It seems that the client didn't specifically say about widget annotations, so that means we can have more flexibility here.

Personally I think we should still keep widget annotations even if it's unchecked, but I will leave the decision to Matt.

@mparizeau-pdftron Can you take a look at the discussion and share your thoughts when you have time? Thanks!

mparizeau-pdftron: Yeah I agree let's keep widgets when this option is checked

JimmyLin39: sounds good, I will modify the code.

src/components/PrintModal/PrintModal.js
```diff
@@ -261,17 +270,17 @@ class PrintModal extends React.PureComponent {
   };
 
   drawAnnotationsOnCanvas = (canvas, pageNumber) => {
-    const annotations = core
+    const widgetAnnotations = core
```
ZhijieZhang, line 273: Good change. Thanks!

PR #685, opened 2020-06-17
Submenu accessibility

liamross:
1. Extracted submenu logic into `FlyoutMenu` component
1. Completed submenu accessibility using `FlyoutMenu` for:
    - `ViewControlsOverlay`
    - `MenuOverlay`
    - `ZoomOverlay`
1. Removed unnecessary style sheets
1. Improved screen reader capabilities in `SearchOverlay`

> For future: Could also add aria to tie button to menu when it's open. Can optimize at a later date.


i18n/translation-en.json
```diff
@@ -38,6 +38,7 @@
     "zoom": "Zoom",
     "zoomIn": "Zoom in",
     "zoomOut": "Zoom out",
+    "zoomControls": "Zoom Controls",
```
liamross, line 41: This is only used as the aria-label for the zoom overlay dropdown button

src/components/Button/Button.js
```diff
@@ -44,7 +49,17 @@ const Button = props => {
...
     title,
     style,
+    ariaLabel,
   } = { ...props, ...customOverrides };
+  const [t] = useTranslation();
+
+  let aLabel = ariaLabel;
+  if (!ariaLabel && title) {
+    aLabel = t(title);
+  }
```
liamross, line 59: Nevermind this is better and still readable:

```suggestion
  const aLabel = ariaLabel || title ? t(title) : undefined;
```

src/components/FlyoutMenu/FlyoutMenu.js
```diff
@@ -0,0 +1,154 @@
...
+const MENUS = [
+  'menuOverlay',
+  'groupOverlay',
+  'viewControlsOverlay',
+  'searchOverlay',
+  'signatureOverlay',
+  'zoomOverlay',
+  'redactionOverlay',
+  'toolStylePopup',
+];
```
liamross, line 24: I just grabbed everything from all 3 of the converted menus, someone might have to check this over.

liamross: Some of these might not exist any more. With this centralized FlyoutMenu component at least it's all in one place so we don't have to maintain 3 individual locations.

src/components/FlyoutMenu/FlyoutMenu.js
```diff
@@ -0,0 +1,154 @@
...
+          const focusable = getFocusableElements();
+
+          if (focusable.length === 0) {
+            return;
+          }
+
+          const focusIndex = findFocusableIndex(focusable, document.activeElement);
+
+          if (e.key === 'ArrowUp') {
+            if (focusIndex <= 0) {
```
ZhijieZhang, line 114: I think the logic here can be simplified to
```js
let nextIndex;
if (e.key === 'ArrowUp') {
  nextIndex = (focusIndex - 1 + length) % length;
} else {
  nextIndex = (focusIndex + 1 + length) % length;
}

focusable[nextIndex].focus();
```
(if we want to be fancy （￣︶￣）↗　 )

src/core/accessibility.js
```diff
@@ -1,3 +1,66 @@
+import { isMac } from '../helpers/device';
```
ZhijieZhang, line 1: Do you think we can move this entire file to `helpers`? I think the `core` folder mostly just contains WebViewer APIs.
Would like to hear your thoughts. We can just leave it as it is.

src/core/accessibility.js
```diff
@@ -1,3 +1,66 @@
...
+
 export function isKeyboardNav() {
   return document.documentElement.getAttribute('data-tabbing') === 'true';
 }
+
+export function findFocusableIndex(elements, toFind) {
```
ZhijieZhang, line 7: Kind of sad that we aren't able to use `findIndex` here :(

src/core/accessibility.js
```diff
@@ -1,3 +1,66 @@
...
+  "undo":                   "Control+Z",
+  "stickyNote":             "N",
+  "strikeout":              "K",
+  "underline":              "U",
+  "zoomIn":                 "Control+=",
+  "zoomOut":                "Control+-",
+  "richText.bold":          "Control+B",
+  "richText.italic":        "Control+I",
+  "richText.underline":     "Control+L",
+  "richText.strikeout":     "Control+K",
```
ZhijieZhang, line 53: I'm not sure if we should move this to `hotkeysManager.js` just to colocate hotkeys all together. I guess my main concern is that if someone else changes the hotkeys in that file, will he know that he needs to change this file as well?

src/core/accessibility.js
ZhijieZhang: It's kind of surprising that we can use `forEach` on a HTML element collection though... :0

i18n/translation-en.json
```diff
@@ -38,13 +38,16 @@
...
```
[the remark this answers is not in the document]
liamross: Yeah it should be, I'll do that

src/components/Button/Button.js
```diff
@@ -44,7 +49,14 @@ const Button = props => {
     className,
     title,
     style,
+    ariaLabel,
   } = { ...props, ...customOverrides };
+  const [t] = useTranslation();
+
+  const aLabel = ariaLabel || title ? t(title) : undefined;
+
+  const shortcutKey = title ? title.slice(title.indexOf('.') + 1) : undefined;
```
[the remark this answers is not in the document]
liamross: Hmmm I'm just copying over the code that was used within Tooltip for displaying the shortcut. In this case, it seems to work fine for buttons that actually have shortcuts (some shortcuts actually use two strings joined by a period hence first index).

Perhaps we should revisit this at some point to simplify though.

src/components/FlyoutMenu/FlyoutMenu.js
liamross: Good call, gonna do that

src/core/accessibility.js
liamross: Yep moved, wasn't sure about the purposes of the folders

src/core/accessibility.js
liamross: Yeah kinda a shame, I will implement a for loop instead of forEach though to make it slightly more efficient.

src/core/accessibility.js
liamross: Sounds good, will move it

PR #686, opened 2020-06-18
fix: wrong signature rotation in the overlay image when deleting another signature

ZhijieZhang:


ZhijieZhang:
This is not an issue in the new UI, so only opening for the legacy UI.

PR #689, opened 2020-06-18
Set custom decrypt function if passed

bollain:
Set custom decrypt function if passed. 
Fixes:
https://support.pdftron.com/a/tickets/14451

### DIFF
--- a/src/apis/loadDocument.js
+++ b/src/apis/loadDocument.js
@@ -10,6 +10,7 @@
  * @param {boolean} [options.withCredentials] Whether or not cross-site requests should be made using credentials.
  * @param {string} [options.cacheKey] A key that will be used for caching the document on WebViewer Server.
  * @param {string} [options.password] A string that will be used to as the password to load a password protected document.
+ * @param {function} [options.decrypt] A custom function to decrypt the document
  * @example
 WebViewer(...)
   .then(function(instance) {
--- a/src/apis/loadDocument.js
+++ b/src/apis/loadDocument.js
@@ -11,6 +11,7 @@
  * @param {string} [options.cacheKey] A key that will be used for caching the document on WebViewer Server.
  * @param {string} [options.password] A string that will be used to as the password to load a password protected document.
  * @param {function} [options.decrypt] A custom function to decrypt the document
+ * @param {object} [options.decryptOptions] An object with with properties `type` for the type of encryption and `p` or the password.
  * @example
 WebViewer(...)
   .then(function(instance) {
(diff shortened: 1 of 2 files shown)


src/apis/loadDocument.js
```diff
@@ -10,6 +10,7 @@
  * @param {boolean} [options.withCredentials] Whether or not cross-site requests should be made using credentials.
  * @param {string} [options.cacheKey] A key that will be used for caching the document on WebViewer Server.
  * @param {string} [options.password] A string that will be used to as the password to load a password protected document.
+ * @param {function} [options.decrypt] A custom function to decrypt the document
```
bollain, line 13: Should I also document the decryptOptions?

src/helpers/loadDocument.js
```diff
@@ -71,10 +71,14 @@ const extractXodOptions = options => {
...
 
   if (options.decryptOptions) {
-    xodOptions.decrypt = window.CoreControls.Encryption.decrypt;
+    xodOptions.decrypt = options.decrypt ? options.decrypt : window.CoreControls.Encryption.decrypt;
     xodOptions.decryptOptions = options.decryptOptions;
   }
 
+  if (options.decrypt) {
+    xodOptions.decrypt = options.decrypt;
+  }
```
bollain, line 80: This may be overkill. Im just not sure if custom decrypt also requires decryptOptions.

ZhijieZhang: I am not sure actually. @mparizeau-pdftron can you take a look when you have time? Thanks.

src/apis/loadDocument.js
ZhijieZhang: Yeah, let's document `decryptOptions` as well.
For the documentation, I think we can just borrow from `createDocument`.

```jsdoc
 * @param {boolean} [options.xodOptions.decrypt] - Function to be called to decrypt a part of the XOD file. For default XOD AES encryption pass CoreControls.Encryption.decrypt.
 * @param {boolean} [options.xodOptions.decryptOptions] -  An object with options for the decryption e.g. {p: "pass", type: "aes"} where is p is the password.
```

I guess the main part here is that we want to mention this is for XOD files only.

mparizeau-pdftron: Yes let's add it. It should be an object with properties `type` for the type of encryption and `p` for the password.

src/helpers/loadDocument.js
mparizeau-pdftron: I think in theory you might have it by itself. If we have this if statement I think we could revert the previous code to `xodOptions.decrypt = window.CoreControls.Encryption.decrypt;` because if there is a custom decrypt function defined then it will override it here.

src/apis/loadDocument.js
ZhijieZhang: yeah oops, the documentation for createDocument doesn't seem to be correct.

bollain:
@mparizeau-pdftron @ZhijieZhang updated as per comments, thanks!

PR #694, opened 2020-06-24
[Feature] Rubber stamp update, new custom stamps feature

sisardor:


ZhijieZhang:
This PR has conflicts now, can yo take a look when you have time?

ZhijieZhang:
@khein-pdftron Can you take a look at this PR when you have time? Thanks!

src/components/CreateStampModal/CreateStampModal.scss
```diff
@@ -0,0 +1,223 @@
...
+      }
+
+      .ColorPalette {
+        width: 110px;
+      }
+    }
+  }
+}
+
+// .CustomStampModalXX {
```
sisardor, line 120: Will remove this

khein-pdftron:
Screen Shot 2020-06-30 at 10 02 42 PM

Do we need this gap? We could make the dropdown grow as you make stamps.

Screen Shot 2020-06-30 at 10 01 35 PM

Making too many stamps overflows. Maybe limit how many you can make like we do with signatures? If we did the above then this wouldn't be a problem.

Screen Shot 2020-06-30 at 10 03 36 PM

The right side of the trash can should line up with the header/tabs. It goes too far to the right.

The font size for 'create new stamp' is 16px on desktop. It should be 13px on desktop and tablet and 16px on mobile.

khein-pdftron:
There is something wrong with the time checkbox. I can't click it. It does not respond. I can only click the text.

On mobile, let's make the 'create new stamp' button have the same style as the 'Add Signature' button. See images below:

Screen Shot 2020-06-30 at 10 12 17 PM
Screen Shot 2020-06-30 at 10 12 45 PM

src/components/RubberStampOverlay/RubberStampOverlay.scss
```diff
@@ -3,43 +3,195 @@
...
+      color: #E1E1E3;
+      cursor: default;
+    }
+
+    &.enabled:hover {
+      background: #17324C;
+      border: 1px solid #17324C;
+      color: #FFFFFF;
+    }
   }
```
khein-pdftron, line 196: This css class (.add-custom-stamp-button) looks dead nothing is using this. Remove it?

src/components/RubberStampOverlay/RubberStampOverlay.scss
```diff
@@ -3,43 +3,195 @@
...
+    cursor: pointer;
+    margin-right: 8px;
+
+    img {
+      max-width: 100%;
+      max-height: 100%;
+    }
+    &:hover {
+      background: var(--popup-button-hover);
+    }
```
khein-pdftron, line 98: can you change this to:
```    
&:not(.add-btn) {
      &:hover {
        background: var(--popup-button-hover);
      }
    }
```
just so it is consistent with the add signature button.

PR #712, opened 2020-06-30
When notes panel disabled

bollain:
dont show in accessibility header

Fixes:
https://support.pdftron.com/a/tickets/14757

If WV is in accessibility mode, when the notes panel is disabled, a skip to notes link is still shown to the user. This removes that link if the notes panel is disabled
image

### DIFF
--- a/src/components/Accessibility/Accessibility.js
+++ b/src/components/Accessibility/Accessibility.js
@@ -9,6 +9,7 @@ import './Accessibility.scss';
 class Accessibility extends React.PureComponent {
   static propTypes = {
     isAccessibleMode: PropTypes.bool,
+    isNotesPanelDisabled: PropTypes.bool,
   }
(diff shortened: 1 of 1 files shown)


ZhijieZhang:
Actually... I'm not sure. @liamross Do we need this one for `master`?

liamross:
I think this might be valuable to have in master for sure. It only shows up on tab start yeah? I would probably have to implement something like this at some point so it might be nice to just bring this in. Is it hidden behind a flag or something? I've never seen it myself when playing around with stuff.

bollain:
@liamross yeah it shows up on start tab - it can be somewhat confusing if the notes panel stuff is disabled but Notes is still a tab-able option. It's not hidden behind a flag, I just instantiated WV with accessibilityMode: true.

PR #713, opened 2020-06-30
Improve search result scrolling performance

jniemin:
PR changes SearchResult list to use react-virtualized like NotesPanel.

Note this PR enables Redux devtools extension for dev mode. This won't work if you access it through iframe, but you can take the iframe url and access it directly and then dev tools work.


src/components/SearchResult/SearchResult.js
```diff
@@ -1,55 +1,168 @@
...
+
+function SearchResultListItem(props) {
+  const { result, currentResultIndex, activeResultIndex, onSearchResultClick } = props;
+  const { ambientStr, resultStrStart, resultStrEnd, resultStr } = result;
+  const textBeforeSearchValue = ambientStr.slice(0, resultStrStart);
+  const searchValue = ambientStr === '' ? resultStr : ambientStr.slice(resultStrStart, resultStrEnd);
+  const textAfterSearchValue = ambientStr.slice(resultStrEnd);
+  return (
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events
+    <div
```
ZhijieZhang, line 53: Hmm I'm not very sure if we just want to ignore linting here...

@liamross when you have time can you take a look here and see if it's ok. Thanks.

jniemin: Original code was giving 2 different linting errors. First one was easy fix, but I wasn't sure if we want to add key events to every search result item, so I disabled for now. Happy to add key events there as well if we have defined way how we are usually doing these

liamross: Yeah generally there are two moves: convert to a button then use my `@include button-reset;` sass style to remove button-specific styling (or at least a bulk of it), or add onKeyDown listeners for <kbd>Space</kbd> and <kbd>Enter</kbd>

jniemin: Converted this to use `<button>`

src/components/SearchResult/SearchResult.scss
```diff
@@ -17,6 +22,10 @@
     box-shadow: none;
   }
 
+  &:focus{
+    outline: none;
+  }
+
```
liamross, line 28: `button-reset` already removes outline when not using keyboard nav, but ensures that it exists when user needs to see it to track their keyboard focus. Removing this here will lower accessibility, recommend removing.

```suggestion
```

src/components/SearchResult/SearchResult.js
```diff
@@ -1,55 +1,166 @@
...
+  translate: PropTypes.func.isRequired,
+  onClickResult: PropTypes.func,
+  pageLabels: PropTypes.arrayOf(PropTypes.any),
+};
+
+function SearchResult(props) {
+  const { width, height, activeResultIndex, noSearchResult, searchResults = [], translate, onClickResult, pageLabels } = props;
+  const cellMeasureCache = React.useMemo(() => {
+    return new CellMeasurerCache({ defaultHeight: 50, fixedWidth: true });
+  }, [width]);
```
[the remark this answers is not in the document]
jniemin: Yes, this is intended. React-virtualized works by converting existing elements to absolute elements and this requires it to calculate every elements (in our case the search list item) height. As height is dependent on width, it expects that width in our case is fixed (they recommend only using either dynamic height or width for performance reason). CellMeasurerCache is used to store element heights, but as our search panel can be extended, we need to recalculate heights based on current width of the search panel. Thus having width as a dependency here

src/components/SearchResult/SearchResult.js
```diff
@@ -1,55 +1,166 @@
...
+              result={result}
+              currentResultIndex={index}
+              activeResultIndex={activeResultIndex}
+              onSearchResultClick={onClickResult}
+            />
+          </div>
+        )}
+      </CellMeasurer>
     );
+  }, [cellMeasureCache, searchResults, activeResultIndex]);
```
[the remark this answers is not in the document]
jniemin: This could be possibility in rare cases. I'll add them

src/components/SearchResult/SearchResult.scss
jniemin: Previous implementation did not have the focus outline. I'll remove this

jniemin:
> The results container also flickers a lot when resizing the panel. The resizing of the container also seems to lag. This doesn't happen on master. Might be the calc width on the search result. Not sure.

I can't see the flicker happening, but the lag is there. I tried with latest master and lag is there, even if I remove the result component all together. Seems to be worst on Chrome and not really noticeable with Firefox, and Safari somewhere in between.

I was not able to get UI pixel perfect with virtualized list as it replaces current content with grid and absolute positioned elements. I was able to get to one pixel off, but changing between with and without scrolls I was not able to get it pixel perfect. React-virtualized container is also defined as `overflow:hidden`, and this causes our drop shadow to cut when it goes all the way to right side of the container.

If you have any ideas how we could get this pixel perfect, I'm happy to try out those.

src/components/SearchResult/SearchResult.scss
jniemin: done

src/components/SearchResult/SearchResult.js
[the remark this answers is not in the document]
jniemin: done

khein-pdftron:
master-search-panel-resize
search-virtualization-search-panel-resize

liamross:
^ regarding @khein-pdftron 's gif, I think perhaps right-aligning it will keep the scroll bar in-line with the right side. I think it's moving due to left-aligned box and lagging in resizing, and that effect will be less pronounced if the scroll bar isn't moving from side to side

jniemin:
After doing quite a bit of research I believe that major reason why this happens comes down to the we need to set the width of the component dynamically on every render (meaning container.style.width=xxx). During render we have quite a bit of logic which causes ui thread to be blocked for sort period of time and that causes frame drops. To make matters worse our SearchPanel also set its width dynamically. This is caused by ResizeBar implementation. Why it wasn't noticeable on previous implementation was because element width was not set dynamically. You can see similar lag on our other resizable panels like thumbnail panel and lag is there even without virtualized result list (when it search results says 'no results')

The way we have things implemented now, I can't really see any fix for this, unless we reconsider the how we implement panels.

@mparizeau-pdftron @khein-pdftron What do you think we should do here? Accept that we do have lag and there is UI issue when resizing search panel, drop using react-virtualized here and keep using current implementation, or try to refactor our resizing panel implementation.

jniemin:
@khein-pdftron I merged your changes, removed width dependency which is not needed anymore, and tested that it works as expected. I think we are ready to merge this

PR #720, opened 2020-07-07
removed css grid stlyling as it didn't work in IE11

achenPDFTron:


khein-pdftron:
Fixed on ie11 with fallbacks to display: grid

PR #723, opened 2020-07-07
New setCustomModal API

jniemin:
New setCustomModal API which allows users to create custom modals


src/redux/selectors/exposedSelectors.js
```diff
@@ -170,6 +172,8 @@ export const getCursorOverlayData = state => state.viewer.cursorOverlay;
 
 export const getOpenElements = state => state.viewer.openElements;
 
+export const getDisabledElements = state => state.viewer.disabledElements;
```
jniemin, line 175: As user is able to add more than one custom modal and they are rendered all at the same time, I needed to expose all disabledElements. Previously we had redux selector only to allow predefined modals one by one.

src/apis/setCustomModal.js
```diff
@@ -0,0 +1,48 @@
+import actions from 'actions';
+
+function generateSetCustomModal(store) {
+
+  /**
+   * Add custom modal element to WebViewer.
```
jniemin, line 6: How dow we handle versioning on our documentation? For now this is only for 7.0

ZhijieZhang:
@khein-pdftron Can you review it when you have time? Thanks.

jniemin:
> I tried the sample code in the documentation. I couldn't close the modal by clicking outside of it. Can we make that the default behaviour? All other modals do that.

I'll add that. I will make that on by default and possibility to opt-out from it using options

jniemin:
@khein-pdftron Custom modal now supports automatic closing with option to opt-out

PR #725, opened 2020-07-08
[Feature] Allow blank and transparent color palettes

awy3:
Allow blanks and transparent color palettes

```
 instance.setColorPalette({
    toolNames: ['AnnotationCreateRectangle'],
    colors: [
   '#272727',
    'transparency',
    '#696969',
    '', //will be empty space
    '#FFFFFF',
    '#B54800',
    ]
  });
```
image

### DIFF
--- a/src/components/ColorPalette/ColorPalette.js
+++ b/src/components/ColorPalette/ColorPalette.js
@@ -76,6 +76,19 @@ class ColorPalette extends React.PureComponent {
     const allowTransparent = !(property === 'TextColor' || property === 'StrokeColor');
 
     const palette = overridePalette2 ||  overridePalette?.[colorMapKey] || overridePalette?.global || this.palette;
+    const hasTransparentPalette = palette.some(bg => bg?.toLowerCase() === 'transparency');
+
+    const transparentIcon = (
+      <svg
+        width="100%"
+        height="100%"
+        className={classNames({
+          transparent: true,
+        })}
+      >
+        <line stroke="#d82e28" x1="0" y1="100%" x2="100%" y2="0" strokeWidth="2" strokeLinecap="round" />
+      </svg>
+    );
 
     return (
       <div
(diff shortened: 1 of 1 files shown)


khein-pdftron:
@awy3 Curious on what the use case for this is? Thanks.

ZhijieZhang:
Acrobat doesn't have transparent color for stroke(border). Andrew can you take a look to see if it's in the spec? And how does transparent border look like in Acrobat?

ZhijieZhang:
Do we still need this PR?

Camo30:
Yes, this PR is needed. With this change I'm able to get the following behavior using the legacy UI: the empty item on the lower left is not clickable while without it is clickable which sets the color to transparent which is not correct.

color_palette

awy3:
I updated the PR with 
```
    if (!allowTransparent) {
      palette = palette.filter(p => p?.toLowerCase() !== 'transparency');
    } 
```

So `transparency` won't show up for border and text

src/components/ColorPalette/ColorPalette.js
```diff
@@ -126,15 +147,9 @@ class ColorPalette extends React.PureComponent {
...
-                <svg
-                  width="100%"
-                  height="100%"
-                  className={classNames({
-                    transparent: true,
-                  })}
-                >
-                  <line stroke="#d82e28" x1="0" y1="100%" x2="100%" y2="0" strokeWidth="2" strokeLinecap="round" />
-                </svg>
+              {
```
ZhijieZhang, line 150: nit: need one more indentation

src/components/ColorPalette/ColorPalette.js
```diff
@@ -86,31 +103,35 @@ class ColorPalette extends React.PureComponent {
...
+              className="cell-container"
+              onClick={() => {
+                this.setColor(bg === 'transparency' ? null : bg);
+              }}
             >
               <div
                 className={classNames({
-                  cell: true,
-                  border: bg.toLowerCase() === '#ffffff',
+                  'cell-outer': true,
```
ZhijieZhang, line 118: Since we are adding special checking for `transparency`, it looks like the code starting from line 134 is not needed anymore? To maintain the original behavior we can just add a `transparency` to the end of `this.palette` array.

awy3: Thanks for pointing this out. I made this change

src/components/ColorPalette/ColorPalette.js
awy3: This section has been removed

PR #737, opened 2020-07-09
Bugfix/mobile measurement overlay

achenPDFTron:
Fixes https://trello.com/c/iPOKvgtb/1148-android-measurement-demo-sample

### DIFF
--- a/src/components/MeasurementOverlay/MeasurementOverlay.js
+++ b/src/components/MeasurementOverlay/MeasurementOverlay.js
@@ -108,8 +108,14 @@ class MeasurementOverlay extends React.PureComponent {
 
   isMouseInsideRect = (e, overlayElement) => {
     const overlayRect = overlayElement.getBoundingClientRect();
-    const { clientX: x, clientY: y } = e;
-
+    let x,y;
+    if (e instanceof TouchEvent && e.touches) {
+      x = e.touches[0].clientX;
+      y = e.touches[0].clientY;
+    } else {
+      x = e.clientX;
+      y = e.clientY;
+    }
     return (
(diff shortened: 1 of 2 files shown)


src/components/MeasurementOverlay/MeasurementOverlay.scss
```diff
@@ -15,7 +15,6 @@
     right: 8px;
     padding: 8px;
     font-size: 14px;
-    z-index: 50;
```
achenPDFTron, line 18: @khein-pdftron as discussed. removed this so that z-index is same for both mobile and non-mobile

src/components/MeasurementOverlay/MeasurementOverlay.js
```diff
@@ -108,8 +108,14 @@ class MeasurementOverlay extends React.PureComponent {
 
   isMouseInsideRect = (e, overlayElement) => {
     const overlayRect = overlayElement.getBoundingClientRect();
-    const { clientX: x, clientY: y } = e;
-
+    let x,y;
+    if (e instanceof TouchEvent && e.touches) {
```
achenPDFTron, line 112: @ZhijieZhang please review this part. it seems, when in mobile, event type is not of type mouse event and is of type touch event. hence that is why the measurement overlay does not fade when you are drawing under it.

ZhijieZhang: I see. This change makes sense and looks good to me.

khein-pdftron: @achenPDFTron Nice job! Verified that the fading works.

PR #738, opened 2020-07-09
feat: updated a11y component

liamross:
1. Simplified a11y component, added jump to search
1. Removed hotkey enter listeners

TO TEST:

1. Run accessibility sample
   - Should have helper to skip to sections when you tab into it
   - Should skip to sections correctly
1. Notes on annotations should be focusable
   - Pressing enter to select a note when tabbed to it should not insert a line break into the comment field

### DIFF
--- a/src/components/Accessibility/Accessibility.scss
+++ b/src/components/Accessibility/Accessibility.scss
@@ -2,37 +2,33 @@
 
 .Accessibility {
   position: fixed;
-  left: 50%;
-  transform: translate(-50%);
-  z-index: 30;
+  z-index: 9999;
 
   display: flex;
+  align-items: center;
   overflow: hidden;
 
   background: white;
-  color: $pdftron-blue;
-  font-style: italic;
+  color: black;
   font-size: 1.2em;
-  border-bottom: 1px solid var(--border);
+  border: 1px solid var(--border);
+  
+  height: $top-bar-height;
+  padding: 0 10px;
 
   &.visible {
-    height: $top-bar-height;
-    padding: 10px;
+    transform: translateY(0%);
   }
 
   &.hidden {
-    height: 0;
-    padding: 0;
+    transform: translateY(-100%);
   }
 
-  div {
-    margin: 2px 7px;
-    padding: 1px;
-  }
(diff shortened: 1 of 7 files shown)


src/helpers/hotkeysManager.js
```diff
@@ -330,19 +330,6 @@ WebViewer(...)
...
-        } else if (document.activeElement.className === 'skip-to-document') {
-          document.getElementById('pageText0').focus();
-        } else if (document.activeElement.className === 'skip-to-notes') {
-          dispatch(actions.openElement('notesPanel'));
-          const noteEl = document.querySelector('.Note');
-          if (noteEl) {
-            noteEl.focus();
-          }
-        }
-      },
```
liamross, line 345: This section wasn't needed. Notes are now focusable and keyboard clickable, and the a11y block elements have their own on-click functions.

src/components/Note/Note.js
```diff
@@ -123,7 +124,14 @@ const Note = ({ annotation }) => {
   );
 
   return (
-    <div ref={containerRef} className={noteClass} onClick={handleNoteClick} onKeyDown={handleNoteKeydown}>
+    <div
+      role="button"
+      tabIndex={0}
```
liamross, line 129: focusable and with a role

src/components/Accessibility/Accessibility.scss
```diff
@@ -2,37 +2,33 @@
 
 .Accessibility {
   position: fixed;
-  left: 50%;
-  transform: translate(-50%);
-  z-index: 30;
+  z-index: 9999;
```
liamross, line 5: 30 wasn't enough to put it on top of webviewer

ZhijieZhang:
I tested and it worked well. Can you take a look at the code in the Escape hotkey? I feel like it's dead.

If I remember correctly the original functionality is that in the a11y mode, when we hit `escape`, the helper will pop up and users can then tab to choose to skip to document or the notes panel.

However that code seems to be dead or not functioning for a long time, so what you have now is great. You can merge it in if Kristian is busy.

PR #744, opened 2020-07-13
[Fix] decrease swipe threshold to make swipe more responsive

jniemin:
Changed swipe threshold to 0.1 of page width from 0.35, which seems to most optimal value for user experience.

See:
https://trello.com/c/oOaKumTk/1108-on-continuous-page-mode-swipe-left-and-right-not-registered-always-ios-1351


jniemin:
> When you have time can you open PRs with the same change to `7.0`, `legacy/master` and `legacy/7.0`? So that this change will be in the 7.0 release.
> 
> You can also just cherry-pick and push to origin directly. Thanks!

PRs done

PR #751, opened 2020-07-13
Used tReady prop to not render component until translation file has b…

achenPDFTron:
Fixes https://trello.com/c/uSOD1bd6/1166-untranslated-text-in-header-before-the-translations-are-loaded

To replicate:
Load the UI, you will see some untranslated text in the header.
Its for a brief moment.

Solution:
Made use of 
https://react.i18next.com/legacy-v9/step-by-step-guide#c-handle-rendering-while-translations-are-not-yet-loaded
https://react.i18next.com/latest/usetranslation-hook#not-using-suspense

So that component doesn't appear until translation file has been loaded

### DIFF
--- a/src/components/ToolsOverlay/ToolsOverlay.js
+++ b/src/components/ToolsOverlay/ToolsOverlay.js
@@ -109,6 +109,7 @@ class ToolsOverlay extends React.PureComponent {
   render() {
     const {
       t,
+      tReady,
       isDisabled,
       isOpen,
       toolNames,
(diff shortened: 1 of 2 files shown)


src/components/ToolsOverlay/ToolsOverlay.js
```diff
@@ -119,7 +120,8 @@ class ToolsOverlay extends React.PureComponent {
     } = this.props;
 
     const isVisible = (isOpen || true) && !isDisabled;
-    if (!isVisible) {
+    // if translation file is not ready, don't render else it looks weird and bad to see untranslated text
+    if (!isVisible || !tReady) {
```
achenPDFTron, line 124: Assuming we are fine with using the `tReady` prop, its possible we may run into future cases like this when we create new components.

I think we should use this at the top level somewhere instead?

khein-pdftron: Yeah we can abstract it in the future. I was thinking we can just hide the text. Instead of disabling the component but this works for now.

PR #760, opened 2020-07-14
[Update] Pt_br lang update

sisardor:


ZhijieZhang:
Does the new UI use this translation as well?
If yes then you need to open one to `master` as well.

PR #765, opened 2020-07-14
Fixed scrollTo not working in IE11

achenPDFTron:
https://trello.com/c/85cmruGj/1162-button-to-scroll-tools-to-the-right-or-left-doesnt-work-in-ie-or-old-edge

SO link that helped me:
https://stackoverflow.com/questions/51517324/scrollto-method-doesnt-work-in-edge


achenPDFTron:
@ZhijieZhang addressed. removed brackets.

PR #766, opened 2020-07-14
Accessibility fixes

liamross:
* fix(a11y): added aria-labels to prev and next buttons in toolbar
* feat(a11y): remove all ids from icons
* feat(a11y): added aria to inputs and buttons

TODO: eventually we should clean up the icons to remove all the unnecessary additional content


liamross:
> Why is having an `id` on svgs bad? Just curious.

In this case they're not used and add unnecessary space, but the official reasoning is [discussed here](https://dequeuniversity.com/rules/axe/3.5/duplicate-id) (basically it should refer to a unique item and duplicates may mess up labeling etc -- obviously not applicable in this case, but that's the reasoning).

PR #776, opened 2020-07-16
[BUGFIX] Zoom in and zoom out steps are now more predictable

jniemin:
Zoom in and out functionality is now more predictable and bit more gradual (more steps). See code comment below for the functionality.

https://trello.com/c/gje4MpLD/1014-weird-increments-decrements-in-zoom-level

```
/**
 * zoomIn works as follows. Every time user zooms in we take current zoom level and compare it to
 * zoom factors map which currently is :
 * {
    '0.075': [null, 0.8],
    '0.25': [0.8, 1.5],
    '0.5': [1.5, 2.5],
    '1': [2.5, 4],
    '2': [4, 8],
    '4': [8, 32],
    '8': [32, 64],
    '16': [64, null],
  }
 * Based on the range we select step size from this map. For example if current zoom level is 110% (1.1)
 * then we can see that it falls to range [0.8, 1.5] and step size for this is 0.25. We add this step
 * to current level so we end up to 1.35. We also store this step to stack so we can follow how after few steps we got there.
 *
 * We need to keep the step history to make our steps predictable in all cases.
 * Consider case where we start from zoom level 140% (1.4) and zoomIn. We would end up to 1.4 + 0.25 = 1.65 (165%).
 * If user would now click zoomOut, we would end up 1.65 - 0.5 = 1.15 (115%) which is not the same 140% where we started.
 * But as we store the step history we do 1.65 - 0.25 (value from step history) and end up to 1.4 (140%).
 */
```

### DIFF
--- a/src/helpers/zoom.js
+++ b/src/helpers/zoom.js
@@ -1,58 +1,28 @@
 import core from 'core';
 import { stepToZoomFactorRangesMap, getMaxZoomLevel, getMinZoomLevel } from 'constants/zoomFactors';
 
-export const zoomIn = () => {
-  const currentZoomFactor = core.getZoom();
-  if (currentZoomFactor === getMaxZoomLevel()) {
-    return;
-  }
-
-  const step = getStep(currentZoomFactor);
-  const newZoomFactor = currentZoomFactor + step;
-  zoomTo(Math.min(newZoomFactor, getMaxZoomLevel()));
-};
-
-export const zoomOut = () => {
-  const currentZoomFactor = core.getZoom();
-  if (currentZoomFactor === getMinZoomLevel()) {
-    return;
-  }
-
-  const step = getStep(currentZoomFactor);
-  const newZoomFactor = currentZoomFactor - step;
-  zoomTo(Math.max(newZoomFactor, getMinZoomLevel()));
-};
-
-const getStep = currentZoomFactor => {
+function getStep(currentZoomFactor) {
   const steps = Object.keys(stepToZoomFactorRangesMap);
   const step = steps.find(step => {
     const zoomFactorRanges = stepToZoomFactorRangesMap[step];
-
     return isCurrentZoomFactorInRange(currentZoomFactor, zoomFactorRanges);
   });
 
   return parseFloat(step);
-};
+}
 
-const isCurrentZoomFactorInRange = (zoomFactor, ranges) => {
-  if (ranges[0] === null) {
-    return zoomFactor <= ranges[1];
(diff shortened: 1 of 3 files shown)


src/helpers/zoom.js
```diff
@@ -1,65 +1,145 @@
...
-    return;
-  }
-
-  const step = getStep(currentZoomFactor);
-  const newZoomFactor = currentZoomFactor - step;
-  zoomTo(Math.max(newZoomFactor, getMinZoomLevel()));
-};
-
-const getStep = currentZoomFactor => {
+function getStep(currentZoomFactor) {
```
achenPDFTron, line 4: It looks like the convention of declaring a function is slightly different from the previous code.

I observe  other .js files in this helpers folder follow different format as well.
i.e cssVariablesParser.js vs. acessibility.js

@ZhijieZhang I think we should have a consistent convention?

jniemin: I think we could consider going back using traditional functions in most cases. Arrow functions are anonymous functions and makes debugging a lot harder. See for example this performance analysis: 
Screen Shot 2020-07-17 at 9 06 53 AM

Would be so much more informative if we would see function names rather than 3 levels of anonymous. 

There are certain cases though where it is okay to use arrow functions like if they are class properties or used in some of the callbacks like `map(()=>{})`. In cases like `getStep` there is no benefit of using arrow function (like saving this binding) instead of normal function definition, only downside of anonymous function.

achenPDFTron: @jniemin  Interesting. Thanks for pointing that out.
I'm fine with that.

@ZhijieZhang what are your thoughts?

ZhijieZhang: I think both styles can exist in the codebase since both are valid JavaScript and in the end, Babel will just transpile arrow functions to normal functions.

When I started working on this project I think we were already be using arrow functions in many places. So I just kept using this, even for functional components. I think I like how arrow functions visually make the code a bit clearer to read, but Jussi makes a good point here about normal(plain?) functions. Given that a lot of the places are using arrow functions, I don't think we should force using `function` at this moment since they are many places to modify. But I don't think we should disallow the usage of `function` either.

I just used to performance analyzer for the UI and it seems that if we use `.map(() => {...})` then it will show up as an `(anonymous)`. If our goal is to make the debugging experience very well, then I guess eventually we want to use `.map(function funcName() {}.bind(..))` everywhere in the code? This feels more verbose than just an arrow function :(

Also I tried in Babel playground and it seems that arrow functions will be transpiled to a function with a name. Not 100% sure but does this mean using arrow functions won't actually make the code harder to debug?

PR #784, opened 2020-07-20
[Bugfix] Update UI to use the updated Print API

zsydyc:
In core, I made a change to replace jQuery print function. 
In this PR [merge to master], I update UI to connect with the `print` function defined inside getAnnotationManager
Once the Core PR get merge in, we can test using a file "Test for Print API.pdf" for testing.

### DIFF
--- a/src/apis/print.js
+++ b/src/apis/print.js
@@ -16,6 +16,8 @@ WebViewer(...)
 import print from 'helpers/print';
 import selectors from 'selectors';
 
-export default store => () => {
-  print(store.dispatch, selectors.isEmbedPrintSupported(store.getState()));
+export default store => {
+  window.docViewer.getAnnotationManager().getFieldManager().setPrintHandler(() => {
+    print(store.dispatch, selectors.isEmbedPrintSupported(store.getState()));
+  });
 };
--- a/src/index.js
+++ b/src/index.js
@@ -21,6 +21,7 @@ import setupLoadAnnotationsFromServer from 'helpers/setupLoadAnnotationsFromServ
 import eventHandler from 'helpers/eventHandler';
 import setupI18n from 'helpers/setupI18n';
 import setAutoSwitch from 'helpers/setAutoSwitch';
+import setPrintHandler from './apis/print';
 import setDefaultDisabledElements from 'helpers/setDefaultDisabledElements';
 import setupDocViewer from 'helpers/setupDocViewer';
 import setDefaultToolStyles from 'helpers/setDefaultToolStyles';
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -142,6 +143,9 @@ if (window.CanvasRenderingContext2D) {
       const tool = docViewer.getTool(window.Tools.ToolNames.STICKY);
       tool?.setSaveViewState(true);
     }
+    docViewer.on('documentLoaded', () => {
```
mparizeau-pdftron, line 146: Do we need to do this on every document load or can we do it just once initially?

src/apis/print.js
```diff
@@ -16,6 +16,8 @@ WebViewer(...)
 import print from 'helpers/print';
 import selectors from 'selectors';
 
-export default store => () => {
-  print(store.dispatch, selectors.isEmbedPrintSupported(store.getState()));
+export default store => {
+  window.docViewer.getAnnotationManager().getFieldManager().setPrintHandler(() => {
```
mparizeau-pdftron, line 20: If we change this file won't this change it so that anyone who calls `instance.print()` won't actually set the print modal, it will just set this function?

src/index.js
zsydyc: My understanding (from what @ZhijieZhang taught me): before we load a new document, we will call `docViewer.closeDocument` to close the current document, and `docViewer.closeDocument` will trigger the following functions call chain:
In file `DocumentViewer.ts` when called `closeDocument()` this will trigger -> `dispose()` in the same file, then trigger -> `annotationManager.dispose()` (in AnnotationManager.js), then trigger -> `this.init()` (in file `FieldManager.js`), and this will set my printHandler to null; 

This is why my puppeteer test failed (since printHandler = null).

src/apis/print.js
zsydyc: I didn't see the old function `instance.print()` ever being used, so I thought we can just replace it. Let's chat in person when @mparizeau-pdftron you have time.

zsydyc: After discussing with Matt, functions under APIs could possible be called by our clients, so I am reverting this function and move `setPrintHandler()` to `src/index.js`

src/index.js
zsydyc: Move setPrintHandler() to here

zsydyc:
Hey @mparizeau-pdftron just want to check if there is anything else you want me to change before merging in this PR.

zsydyc:
> Let's do what Zhijie suggested and move to onDocumentLoaded

This has been done!

PR #787, opened 2020-07-20
[Bugfix] [merge to legacy/7.0] Update UI to use the updated Print API

zsydyc:
In core, I made a change to replace jQuery print function. 
In this PR [merge to legacy/7.0], I update UI to connect with the `print` function defined inside getAnnotationManager
Once the Core PR get merge in, we can test using a file "Test for Print API.pdf" for testing.


zsydyc:
we don't need this feature in 7.0, so I am closing this PR

PR #788, opened 2020-07-20
Able to persist snap mode checkbox, put in i18n for enable snapping a…

achenPDFTron:
Fixes https://trello.com/c/IdoxhZx7/1175-enable-snapping-for-measurments-option-seems-broken-no-translation-option-unchecks-itself-when-closing-menu-its-onchange-has-a-c

Added i18n for "Enable Snapping"
Persist checkbox value
Hide checkbox for ellipse annot


i18n/translation-en.json
```diff
@@ -267,7 +267,8 @@
     },
     "shared": {
       "page": "Page",
-      "precision": "Precision"
+      "precision": "Precision",
+      "enableSnapping": "Enable snapping for measurement tools"
```
achenPDFTron, line 271: Let me know if the label should be changed.
didn't want to say "all measurement tools" because ellipse doesn't have snap mode

src/redux/actions/internalActions.js
```diff
@@ -373,4 +373,9 @@ export const setIsProgrammaticSearchFull = isProgrammaticSearchFull => ({
 export const setNoteTransformFunction = noteTransformFunction => ({
   type: 'SET_NOTE_TRANSFORM_FUNCTION',
   payload: { noteTransformFunction },
+});
+
+export const setEnableSnapMode = enable => ({
+  type: 'SET_ENABLE_SNAP_MODE',
```
achenPDFTron, line 379: i think this is my first time setting up a redux state in this code base. let me know if i put it incorrectly

PR #801, opened 2020-07-27
Refactor link modal to improve the performance of documents with many pages

mparizeau-pdftron:
The link modal creates a dropdown with an option for every page number
so for large documents (thousands of pages) this can take quite a while
to render.

This change updates the behavior so that you can type the page number
instead and the link button is disabled if the page is out of bounds.

### DIFF
--- a/src/components/LinkModal/LinkModal.js
+++ b/src/components/LinkModal/LinkModal.js
@@ -180,18 +180,6 @@ const LinkModal = () => {
     }
   }, [tabSelected, isOpen, pageNumberInput, urlInput]);
 
-  const setDropdownNumbers = () => {
-    const numbers = [];
-    for (let i = 1; i <= totalPages; i++) {
-      numbers.push(
-        <option key={i} value={i}>
-          {i}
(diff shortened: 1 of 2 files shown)


mparizeau-pdftron:
Yeah I'm just working on the PR for the new UI, will be up soon. Switching from dropdown to text input could work when there are more pages, I was just thinking of keeping it simpler by doing it this way.

src/components/LinkModal/LinkModal.js
```diff
@@ -240,18 +228,19 @@ const LinkModal = () => {
...
-              </select>
+                onChange={e => setPageNumber(parseInt(e.target.value, 10))}
+                min={1}
+                max={totalPages}
+              />
               <Button
                 dataElement="linkSubmitButton"
                 label={t('action.link')}
                 onClick={addPageLink}
+                disable={pageNumber < 0 || pageNumber > totalPages}
```
ZhijieZhang, line 243: Same comments here as the other PR.

PR #803, opened 2020-07-28
Refactor link modal to improve the performance of documents with many pages

mparizeau-pdftron:
The link modal creates a dropdown with an option for every page number
so for large documents (thousands of pages) this can take quite a while
to render.

This change updates the behavior so that you can type the page number
instead and the link button is disabled if the page is out of bounds.

### DIFF
--- a/src/components/LinkModal/LinkModal.js
+++ b/src/components/LinkModal/LinkModal.js
@@ -195,18 +195,6 @@ const LinkModal = () => {
     }
   }, [tabSelected, isOpen, pageNumberInput, urlInput]);
 
-  const setDropdownNumbers = () => {
-    const numbers = [];
-    for (let i = 1; i <= totalPages; i++) {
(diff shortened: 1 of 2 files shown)


src/components/LinkModal/LinkModal.js
```diff
@@ -261,18 +249,19 @@ const LinkModal = () => {
...
-                  </select>
+                    onChange={e => setPageNumber(parseInt(e.target.value, 10))}
+                    min={1}
+                    max={totalPages}
+                  />
                   <Button
                     dataElement="linkSubmitButton"
                     label={t('action.link')}
                     onClick={addPageLink}
+                    disabled={pageNumber < 0 || pageNumber > totalPages}
```
ZhijieZhang, line 264: Should the condition be `pageNumber < 1` instead?

PR #804, opened 2020-07-28
[Feature] Add support for text editing

zsydyc:
Add text editing features.

There are some known issues from my testing
1. After the text is replaced, there will be some alignment issues with the new text. The issue seems to come from C`ontentReplacer.addText()` and it happens to all the platforms.
2. If the updated text length is longer than the page maximum width,  then the part that exceeds the page will not be displayed. (Also seems to happen on all the platforms)
3. The textarea is reusing NoteTextarea component, which doesn't allow the user to select words using the mouse cursor, I don't know this is intended.


src/components/EditTextModal/EditTextModal.js
```diff
@@ -0,0 +1,160 @@
...
+
+        // clear the cache
+        window.docViewer.refreshAll();
+        // update viewer with new document
+        window.docViewer.updateView();
+        // Annotations may contain text so we need to regenerate
+        // our text representation
+        window.docViewer.getDocument().refreshTextData();
+      }
+      // doc.unlock();
```
zsydyc, line 110: For some reason, this unlock gives me a warning:
`[WARNING], the object to be unlocked was not found in the unlock list. Unlocking may cause errors.`

I am not sure what's wrong with it

src/components/TextPopup/TextPopup.js
```diff
@@ -65,11 +68,21 @@ const TextPopup = () => {
...
-          img="ic_copy_black_24px"
-          onClick={copyText}
+          dataElement="editTextButton"
+          title="action.edit"
+          img="ic_edit_black_24px"
+          onClick={() =>
+            dispatch(actions.openElement('editTextModal'))
+          }
         />
+        {fullAPI && (
```
ZhijieZhang, line 78: I think we always want to show the copy text button, but only show the edit text button when fullAPI is enabled.

ZhijieZhang:
UI related code looks good to me. It's consistent with other modal components. 

I can't comment on the main logic for replacing text, @mparizeau-pdftron can you take a look when you have time?

@liamross When you have time can you take a look and see if there are a11y changes you'd like to have for this component?

liamross:
I'm going to push through some changes. The main changes are the following:

1. Added the modal to the hotkeys list so that escape closes it
1. Added a FocusTrap around the modal to trap tab internally
1. Made the buttons actual buttons (this is important in order to ensure they are focusable. In the future, make any buttons a `button` html element. There is a sass mixin `@include button-reset;` which will remove button styling and add a focus style, allowing you to add the same css you normally would)
1. Removed unused ref (also, you can use `useRef` instead of `createRef` inside of functional components -- also I fixed the broken `AutoResizeTextarea` that was causing you to need to put the ref there in the first place)
1. Made button style common between the two buttons (they are mostly the same, so better to share common style -- although I know the style was copied from `EditingBookmark` so no worries)

Also, unrelated to your changes but related enough to add some a11y to:

1. Made the `TextPopup` allow arrow keys to focus individual elements (the thing that appears when you focus text -- so kinda related)
1. Turn on visual focusing when using arrow keys to navigate (since `TextPopup` isn't findable with tab, only by selecting text, had to make it so the arrows turned on visual focusing)
1. Made print modal only select page range when actual value was entered (stop it from selecting that radio button simply from focusing the input field)

src/components/EditTextModal/EditTextModal.js
```diff
@@ -0,0 +1,156 @@
...
+    closed: !isOpen,
+  });
+
+  return isDisabled ? null : (
+    <Swipeable onSwipedUp={closeModal} onSwipedDown={closeModal} preventDefaultTouchmoveEvent>
+      <FocusTrap locked={isOpen}>
+        <div className={modalClass} data-element="editTextModal" onMouseDown={closeModal}>
+          <div className="container" onMouseDown={e => e.stopPropagation()}>
+            <div className="swipe-indicator" />
+            <p>Enter the text you want to replaced</p>
```
liamross, line 139: This should be added to the translation file (and then run the translation script to generate the translation across all languages)

src/components/EditTextModal/EditTextModal.js
```diff
@@ -0,0 +1,156 @@
...
+  useEffect(
+    () => {
+      if (isOpen) {
+        //  prepopulate selected text
+        setNewText(core.getSelectedText());
+        // convert and store selected text coordinates
+        storeConvertedPDFCoordinates();
+      }
+    },
+    [isOpen]
```
jniemin, line 45: It's good practice to have possible changing functions as dependency as well. So in this case `[isOpen, setNewText]`.

`storeConvertedPDFCoordinates` is better to define inside `useEffect` as it is only used there. Otherwise you need to set it as dependency and then you also need to make sure it's not change on every re-render. If you move it inside useEffect you don't have to worry about that.

src/components/EditTextModal/EditTextModal.js
```diff
@@ -0,0 +1,156 @@
...
+
+import defaultTool from 'constants/defaultTool';
+import core from 'core';
+import actions from 'actions';
+import selectors from 'selectors';
+import AutoResizeTextarea from 'components/NoteTextarea/AutoResizeTextarea';
+
+import './EditTextModal.scss';
+
+const EditTextModal = () => {
```
jniemin, line 16: I think we should use normal functions rather than arrow functions when defining component. This make sure that in all dev tooling and debugging components will have names and they are not anonymous. React and babel might handle convert this, but using normal function makes sure that it is defined as named function rather than anonymous

src/components/TextPopup/TextPopup.js
```diff
@@ -61,9 +69,19 @@ const TextPopup = () => {
...
+      onClick={onClose}
     >
       <CustomizablePopup dataElement="textPopup">
+        {fullAPI && (
+          <ActionButton
+            dataElement="editTextButton"
+            title="action.edit"
+            img="ic_edit_black_24px"
+            onClick={() =>
+              dispatch(actions.openElement('editTextModal'))
```
jniemin, line 81: Wrap this onClick to `useCallback` so it doesn't cause re-render of ActionButton on every render

src/components/EditTextModal/EditTextModal.js
```diff
@@ -0,0 +1,160 @@
...
+  const [isDisabled, isOpen] = useSelector(state => [
+    selectors.isElementDisabled(state, 'editTextModal'),
+    selectors.isElementOpen(state, 'editTextModal'),
+  ]);
+
+  const [t] = useTranslation();
```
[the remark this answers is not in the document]
liamross: Just searching, it's done more commonly with `[t]` (27 results) than `{ t }` (6 results). Obviously we should decide on one though.

src/components/EditTextModal/EditTextModal.js
liamross: You don't have to add `setNewText` since React recognizes that it's a useState setter (if you use the react hooks eslint plugin, it will ignore `setNewText`

src/components/EditTextModal/EditTextModal.js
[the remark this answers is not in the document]
jniemin: I stand corrected :D

src/components/EditTextModal/EditTextModal.js
jniemin: Sure, that is the case. I just like to be more explicit and not leave things for the reader to guess which ones are implicitly handled.

Anyways `storeConvertedPDFCoordinates` should be still moved inside the effect

src/components/EditTextModal/EditTextModal.js
zsydyc: The translation has been added

zsydyc:
> accessibility changes

Thanks @liamross, the accessibility changes you made are awesome.

zsydyc:
> UI issue:
> Too wide in some widths for tablet mode.
> The text: "Enter the text you want replaced" should be font size 15px. And the cancel / save buttons should be aligned to the right.



> ter the text you want replaced" should be font size 15px. And the ca

Hey @khein-pdftron Thanks for the styling suggestions.
I have adjusted the text size to 15px, and right aligned the buttons.
For the width for the text editing area, I noticed the signature area is around 480px width, which is too narrow to hold the entire one line text content. (See screenshot below) 
image
Since we use line break `\n` to detect the actual content need to be updated, I want the content to be displayed in one line, so I want to set the width to 650px (which is slightly bigger, but will also fit for tablet)

src/components/EditTextModal/EditTextModal.js
zsydyc: Thanks for the suggestion, I have moved `storeConvertedPDFCoordinates` to the `useEffect` hook

liamross:
I think it might be ok to have the line breaks and keep the modal narrow, since we will never know for sure how wide the PDF content will be anyways and how many line breaks there will be. @kristianhein what do you think? Should it be narrow regardless of the line breaks?

src/components/EditTextModal/EditTextModal.js
```diff
@@ -0,0 +1,158 @@
...
+    window.PDFNet.runWithCleanup(async() => {
+      const doc = await window.docViewer.getDocument().getPDFDoc();
+      // Treat each line as a rectangle region and update them separately
+      const lines = newText.split(/\r?\n/);
+      doc.initSecurityHandler();
+      doc.lock();
+      const PDFNet = window.PDFNet;
+      // Note: If the number of lines are more than number of rects defined in textCoordinates, then we just ignore the rest
+      if (textCoordinates && textCoordinates.length > 0) {
+        for (let i=0; i<textCoordinates.length; i++) {
```
mparizeau-pdftron, line 92: Character spacing should be like this? `let i = 0; i < textCoordinates.length; i++)`

This isn't in our linting currently?

src/components/EditTextModal/EditTextModal.js
mparizeau-pdftron: I think runWithCleanup will automatically unlock for you https://www.pdftron.com/documentation/web/guides/full-api/advanced-features/#full-api-locking

So you can just delete that line

src/components/EditTextModal/EditTextModal.js
liamross: It might be in our linting, but unless devs have eslint integrated in their IDEs they might miss it

zsydyc: I have eslint integrated with my IDE this section of code has been formatted.

zsydyc:
I made another change to make this text editing feature to be disabled by default, but can all  `instance.enableFeature` to turn on

src/redux/actions/internalActions.js
```diff
@@ -242,6 +242,10 @@ export const setMouseWheelZoom = (enableMouseWheelZoom = true) => ({
   type: 'SET_MOUSE_WHEEL_ZOOM',
   payload: { enableMouseWheelZoom },
 });
+export const setTextEdit = (enableTextEdit = true) => ({
```
ZhijieZhang, line 245: Similar to my comment at https://github.com/PDFTron/webviewer-ui/pull/824#discussion_r474273242.

zsydyc: Same as the other comment, removed TextEdit reducer and actions

zsydyc:
> There're also some conflicts that we need to resolve.

merge conflict resolved, there will be some extra conflict after we merge: https://github.com/PDFTron/webviewer-ui/pull/824

src/components/TextPopup/TextPopup.js
```diff
@@ -61,9 +76,17 @@ const TextPopup = () => {
       data-element="textPopup"
       ref={popupRef}
       style={{ ...position }}
-      onClick={() => dispatch(actions.closeElement('textPopup'))}
+      onClick={onClose}
     >
       <CustomizablePopup dataElement="textPopup">
+        {fullAPI && isTextEditEnabled && (
```
ZhijieZhang, line 82: It looks like `isTextEditEnabled` is always false so there's no way to turn it on. As I mentioned in another comment(or during screenshare?) we don't need to have the extra state in the store to handle toggling. The `ActionButton` component renders a `Button` component, which handles `instance.disableElements`.

ZhijieZhang:

PR #805, opened 2020-07-28
[Refactor] Search source of truth in core

jniemin:
### Description

So far we have had shared responsibility of keeping search state between core apis and redux store. This PR refactor that core is the source of truth for search status.

This PR is depend on webviewer PR (https://github.com/XodoDocs/webviewer/pull/1529) which should be merged first.

### To verify

Open search panel. Do search. Functionality should be still the same

### Related issues

Fixes https://trello.com/c/thaRp9ic/1038-use-activesearchresultchanged-event-in-ui

https://github.com/XodoDocs/webviewer/pull/1529


jniemin:
Wrong branch. Closing

PR #807, opened 2020-07-28
[6.3] [Bugfix] Fixed issue where thumbnail panel was the wrong size after full screen

S0ulDrag0n:
The issue occurs when you load a document with enough pages to get a scroll bar in the page thumbnails panel, open the page thumbnails panel, full-screen WebViewer, and then exit full-screen. The inner thumbnail panel will be slightly longer than expected and the bottom of the panel will be offscreen. The update of the size is handled by the `Measure` component, but it will not trigger `onResize` since the panel did not resize when exiting full-screen.

This PR makes the component listen to the window resize event and updates the size using the container size which will be correct when exiting full-screen.

### DIFF
--- a/src/components/ThumbnailsPanel/ThumbnailsPanel.js
+++ b/src/components/ThumbnailsPanel/ThumbnailsPanel.js
@@ -84,10 +84,12 @@ class ThumbnailsPanel extends React.PureComponent {
   }
 
   onWindowResize = () => {
-    this.setState({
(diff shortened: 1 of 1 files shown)


ZhijieZhang:
Removing myself and adding Andrew to be the reviewer.

S0ulDrag0n:
Yea, I double-checked master and 7.0. The issue does not occur there. Probably due to a large refactoring.

PR #821, opened 2020-07-31
[Feature] Added options object to print API

andreysaf:
Added options to print API:
- All pages
- pages to print,
- include comments
- include annotations

Tested with API, print through UI, print through cmd+P using the built version.

### DIFF
--- a/src/apis/print.js
+++ b/src/apis/print.js
@@ -1,6 +1,10 @@
 /**
  * Print the current document.
  * @method WebViewerInstance#print
+ * @param {boolean} [allPages] If true, all pages will be printed
+ * @param {Array.<number>} [pagesToPrint] If allPages is false and pagesToPrint is passed in, will print only the pages passed in
+ * @param {boolean} [includeAnnotations] If true, will print the documents with the annotations
+ * @param {boolean} [includeComments] If true, will append comments to the document printed
  * @example
 WebViewer(...)
   .then(function(instance) {
@@ -9,6 +13,7 @@ WebViewer(...)
     // you must have a document loaded when calling this api
     docViewer.on('documentLoaded', function() {
       instance.print();
+      // instance.print({ includeComments:true, includeAnnotations: true });
     });
   });
  */
@@ -16,6 +21,24 @@ WebViewer(...)
 import { print } from 'helpers/print';
 import selectors from 'selectors';
 
-export default store => () => {
-  print(store.dispatch, selectors.isEmbedPrintSupported(store.getState()));
+export default store => options => {
+  const defaultOptions = {
+    allPages: false,
+    pagesToPrint: [],
+    includeAnnotations: false,
+    includeComments: false,
+  };
+
+  const printOptions = {
+    ...defaultOptions,
+    ...options,
+  };
+
+  print(
+    store.dispatch,
+    selectors.isEmbedPrintSupported(store.getState()),
+    selectors.getSortStrategy(store.getState()),
+    selectors.getColorMap(store.getState()),
+    printOptions,
+  );
 };
--- a/src/apis/print.js
+++ b/src/apis/print.js
@@ -1,10 +1,6 @@
 /**
  * Print the current document.
  * @method WebViewerInstance#print
- * @param {boolean} [allPages] If true, all pages will be printed
- * @param {Array.<number>} [pagesToPrint] If allPages is false and pagesToPrint is passed in, will print only the pages passed in
- * @param {boolean} [includeAnnotations] If true, will print the documents with the annotations
- * @param {boolean} [includeComments] If true, will append comments to the document printed
  * @example
 WebViewer(...)
   .then(function(instance) {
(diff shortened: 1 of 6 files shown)


src/helpers/hotkeysManager.js
```diff
@@ -6,7 +6,7 @@ import openFilePicker from 'helpers/openFilePicker';
 import copyText from 'helpers/copyText';
 import setToolModeAndGroup from 'helpers/setToolModeAndGroup';
 import { zoomIn, zoomOut } from 'helpers/zoom';
-import { print } from 'helpers/print';
+import print from 'src/apis/print';
```
ZhijieZhang, line 9: Our convention is that the internal code shouldn't use any exposed APIs.
So I think it'd be better if we still use
```
import { print } from 'helpers/print';
```
and then pass all the needed state into this helper function

src/components/MenuOverlay/MenuOverlay.js
```diff
@@ -6,11 +6,11 @@ import core from 'core';
 import { isIOS, isIE } from 'helpers/device';
 import downloadPdf from 'helpers/downloadPdf';
 import openFilePicker from 'helpers/openFilePicker';
-import { print } from 'helpers/print';
+import print from 'src/apis/print';
```
ZhijieZhang, line 9: Same here, see comments in `hotkeysManager.js`.

src/apis/print.js
```diff
@@ -9,13 +13,32 @@ WebViewer(...)
...
 import selectors from 'selectors';
 
-export default store => () => {
-  print(store.dispatch, selectors.isEmbedPrintSupported(store.getState()));
+export default store => options => {
+  const defaultOptions = {
+    allPages: false,
+    pagesToPrint: [],
+    includeAnnotations: false,
+    includeComments: false,
```
ZhijieZhang, line 29: Should `allPages` be `true` and `includeAnnotations` be `true` to match the default options in the print modal?

src/apis/print.js
```diff
@@ -1,6 +1,10 @@
 /**
  * Print the current document.
  * @method WebViewerInstance#print
+ * @param {boolean} [allPages] If true, all pages will be printed
+ * @param {Array.<number>} [pagesToPrint] If allPages is false and pagesToPrint is passed in, will print only the pages passed in
+ * @param {boolean} [includeAnnotations] If true, will print the documents with the annotations
+ * @param {boolean} [includeComments] If true, will append comments to the document printed
```
ZhijieZhang, line 7: I think it'd be good to document what the options will be. One way you can do this is
```
{boolean} [allPages=false]
```
or just have something like `Default is false` at the end of line

src/helpers/print.js
```diff
@@ -32,6 +35,29 @@ export const print = (dispatch, isEmbedPrintSupported) => {
     printPdf().then(() => {
       dispatch(actions.closeElement('loadingModal'));
     });
+  } else if (allPages || includeAnnotations || includeComments || pagesToPrint.length > 0) {
+    if (allPages || pagesToPrint.length === 0) {
```
ZhijieZhang, line 39: I feel like it's more natural to not print any pages if users call

```
instance.print({
  allPages: false,
  includeAnnotations: true,
  pagesToPrint: []
})
```

mparizeau-pdftron:
What about creating a new function called something like `instance.printInBackground` that would just be doing the programmatic print, then we can leave `instance.print` as is? In the future we could consider merging them or renaming them to make more sense as part of a major release.

andreysaf:
@mparizeau-pdftron 

- added a new API called instance.printInBackground(options)
- left instance.print() same

@ZhijieZhang 

- added defaults to the API
- defaults are all false, or blank unless the user provides them in the API
- using helpers instead of external APIs
- added a console.warn for trying to call allPages=false and pagesToPrint=[]

src/apis/printInBackground.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Print the current document.
+ * @method WebViewerInstance#printInBackground
+ * @param {boolean} [allPages=false] If true, all pages will be printed
+ * @param {Array.<number>} [pagesToPrint=[]] If allPages is false and pagesToPrint is passed in, will print only the pages passed in
+ * @param {boolean} [includeAnnotations=false] If true, will print the documents with the annotations
+ * @param {boolean} [includeComments=false] If true, will append comments to the document printed
```
ZhijieZhang, line 7: All the 4 options are optional but if I just call `instance.printInBackground()` I will see
```
webviewer-ui.min.js:50 No pages were passed in to be printed. Please pass either allPages set to true or a range of pages in pagesToPrint.
```
in the console. I think this essentially means that `allPages: true` or `pagesToPrint: ...` is a required option to use this API, which contradicts the documentation.

I think we should either change the documentation here or change the defaultOptions in the code.

src/apis/printInBackground.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Print the current document.
```
ZhijieZhang, line 2: I suggest we explain a little bit about the difference between `printInBackground` and `print`. In a user's perspertive I think it can be confusing for seeing two different APIs with the same description.

andreysaf:
@ZhijieZhang added better docs, removed the check for pagesToPrint being zero, adjusted the required params

PR #822, opened 2020-08-04
[Feature] Sort by statuses

andreysaf:
You are now able to sort by statuses.

image

### DIFF
--- a/i18n/translation-en.json
+++ b/i18n/translation-en.json
@@ -162,7 +162,8 @@
   "option": {
     "notesOrder": {
       "position": "Position",
-      "time": "Time"
+      "time": "Time",
+      "status": "Status"
     },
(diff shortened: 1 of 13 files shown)


andreysaf:
I will add few more sorts.

andreysaf:
Added Author and Type
image

i18n/translation-en.json
```diff
@@ -162,7 +162,10 @@
   "option": {
     "notesOrder": {
       "position": "Position",
-      "time": "Time"
+      "time": "Time",
```
ZhijieZhang, line 165: Can you follow https://github.com/XodoDocs/webviewer/wiki/WebViewer-UI-Localization to add translations for the new text?

PR #823, opened 2020-08-05
[Feature] Added filtering functionality to UI for annots

andreysaf:
You can now filter out annotations through UI:

filter

### DIFF
--- a/src/components/FilterAnnotMondal/FilterAnnotModal.js
+++ b/src/components/FilterAnnotMondal/FilterAnnotModal.js
@@ -84,7 +84,7 @@ const FilterAnnotModal = () => {
 
   const modalClass = classNames({
     Modal: true,
-    LinkModal: true,
+    FilterAnnotModal: true,
     open: isOpen,
     closed: !isOpen,
   });
--- a/src/components/FilterAnnotMondal/FilterAnnotModal.js
+++ b/src/components/FilterAnnotMondal/FilterAnnotModal.js
@@ -67,8 +67,8 @@ const FilterAnnotModal = () => {
     const annots = core.getAnnotationsList();
     // set is a great way to remove any duplicate additions and ensure the unique items are present
     // the only gotcha that it should not be used by state since not always it will trigger a rerender
-    let authorsToBeAdded = new Set();
-    let annotTypesToBeAdded = new Set();
+    const authorsToBeAdded = new Set();
+    const annotTypesToBeAdded = new Set();
     annots.forEach(annot => {
       if (core.getDisplayAuthor(annot) && core.getDisplayAuthor(annot) !== '') {
         authorsToBeAdded.add(core.getDisplayAuthor(annot));
@@ -91,8 +91,8 @@ const FilterAnnotModal = () => {
 
   return isDisabled ? null : (
     <Swipeable onSwipedUp={closeModal} onSwipedDown={closeModal} preventDefaultTouchmoveEvent>
-      <FocusTrap locked={isOpen}>
-        <div className={modalClass} data-element="filterModal" onMouseDown={closeModal}>
+      <div className={modalClass} data-element="filterModal" onMouseDown={closeModal}>
+        <FocusTrap locked={isOpen} focusLastOnUnlock>
           <div className="container" onMouseDown={e => e.stopPropagation()}>
             {core.getAnnotationsList().length > 0 ? (
               <div className="filter-modal">
@@ -163,8 +163,8 @@ const FilterAnnotModal = () => {
               </div>
             )}
           </div>
-        </div>
-      </FocusTrap>
+        </FocusTrap>
+      </div>
     </Swipeable>
   );
 };
--- a/src/components/FilterAnnotMondal/FilterAnnotModal.js
+++ b/src/components/FilterAnnotMondal/FilterAnnotModal.js
@@ -55,7 +55,6 @@ const FilterAnnotModal = () => {
     );
     setAuthorFilter([]);
     setTypesFilter([]);
-    closeModal();
   };
 
   const closeModal = () => {
--- a/src/components/FilterAnnotMondal/FilterAnnotModal.js
+++ b/src/components/FilterAnnotMondal/FilterAnnotModal.js
@@ -65,6 +65,8 @@ const FilterAnnotModal = () => {
 
   useEffect(() => {
     const annots = core.getAnnotationsList();
+    // set is a great way to remove any duplicate additions and ensure the unique items are present
+    // the only gotcha that it should not be used by state since not always it will trigger a rerender
     let authorsToBeAdded = new Set();
     let annotTypesToBeAdded = new Set();
     annots.forEach(annot => {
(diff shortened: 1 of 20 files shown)


src/components/FilterAnnotMondal/FilterAnnotModal.js
```diff
@@ -0,0 +1,152 @@
...
+    core.setToolMode(defaultTool);
+  };
+
+  useEffect(() => {
+    const annots = core.getAnnotationsList();
+    let authorsToBeAdded = new Set();
+    let annotTypesToBeAdded = new Set();
+    annots.forEach(annot => {
+      authorsToBeAdded.add(core.getDisplayAuthor(annot));
+      annotTypesToBeAdded.add(annot.Subject.toLowerCase().replace(/\s/g, ''));
```
ZhijieZhang, line 63: I found that Link annotations don't have `Subject`, so opening the filter modal when a document has link annotations will blow up the app.

src/components/NotesPanel/NotesPanel.js
```diff
@@ -361,6 +318,12 @@ const NotesPanel = () => {
                 </div>
                 <div className="divider" />
                 <div className="sort-row">
+                  <Button
+                    dataElement="filterAnnotationButton"
+                    className="filter-annotation-button"
+                    label={t('component.filter')}
+                    onClick={() => dispatch(actions.openElement('filterModal'))}
```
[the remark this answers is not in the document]
andreysaf: Correct, since the rest is just formatting.

mparizeau-pdftron:
@ZhijieZhang For your comment 2 I think it would probably be difficult to always match the custom function to an author or type because the function could in theory have anything in side of it. As long as we have a data-element for the filter button they could hide that filter and use their own custom filter?

andreysaf:
@ZhijieZhang 

- unchecking filters now shows previously filtered annots
- blank authors or subjects do not show up
- check for translation and if not use the original subject
- modal can be dismissed by esc key
- modal now focus traps the user into making selections
- added comments and resolved all nit

khein-pdftron:
> I didn't review the CSS part, not sure if @khein-pdftron will have some comments on them?
> 
> Other than the comments below, I have two concerns that I think should be addressed.
> 
> 1. Check the box and click the apply button filters the notes in the panel. This is working as expected. But later if I uncheck the same box, and click the apply button, the previous filtered notes don't come back in the panel.
>    I think in this case previous filtered nodes should come back since the box is unchecked.
>    It looks like the current way is to click the `cancel` button is to reset it, but I feel like users also expect that unchecking the box should work.
> 2. `setCustomNoteFilter` is an external API that can be used by our clients. If a client calls `instance.setCustomNoteFilter(annot => annot.Subject !== 'abc')` then annotations with `abc` Subject will not show in the notes panel. But if I open the filter modal, `abc` is not checked. I think in this case `abc` should be checked because it is filtered by the API call.

Thanks @ZhijieZhang. I left Andrey a review for the styling on Xodo. I will approve when those changes are done.

src/components/FilterAnnotMondal/FilterAnnotModal.js
```diff
@@ -0,0 +1,172 @@
...
+  const modalClass = classNames({
+    Modal: true,
+    FilterAnnotModal: true,
+    open: isOpen,
+    closed: !isOpen,
+  });
+
+  return isDisabled ? null : (
+    <Swipeable onSwipedUp={closeModal} onSwipedDown={closeModal} preventDefaultTouchmoveEvent>
+      <FocusTrap locked={isOpen}>
```
liamross, line 94: Looks fantastic! Only suggestion would be to re-focus the "Filter" button when the modal closes (there's a prop for that)

```suggestion
      <FocusTrap locked={isOpen} focusLastOnUnlock>
```

liamross: It does look like that only fixes it if you use escape to close though, let me look into that quickly

src/components/FilterAnnotMondal/FilterAnnotModal.js
```diff
@@ -0,0 +1,172 @@
...
+  const filterApply = () => {
+    dispatch(
+      actions.setCustomNoteFilter(annot => {
+        let type = true;
+        let author = true;
+        if (typesFilter.length > 0) {
+          type = typesFilter.includes(annot.Subject.toLowerCase().replace(/\s/g, ''));
+        }
+        if (authorFilter.length > 0) {
+          author = authorFilter.includes(annot.Author);
```
ZhijieZhang, line 42: Hmm it looks like this one hasn't been addressed.
```
Should we use the display author for authorFilter.includes? I see that authors contains display authors.
```

src/components/FilterAnnotMondal/FilterAnnotModal.js
```diff
@@ -0,0 +1,172 @@
...
+    // set is a great way to remove any duplicate additions and ensure the unique items are present
+    // the only gotcha that it should not be used by state since not always it will trigger a rerender
+    const authorsToBeAdded = new Set();
+    const annotTypesToBeAdded = new Set();
+    annots.forEach(annot => {
+      if (core.getDisplayAuthor(annot) && core.getDisplayAuthor(annot) !== '') {
+        authorsToBeAdded.add(core.getDisplayAuthor(annot));
+      }
+      if (annot.Subject && annot.Subject !== '') {
+        // we want to lowercase it and remove spaces to get i18n translations
```
ZhijieZhang, line 77: Thanks for putting the comment. Now I'm wondering if we need to display the translated text?

For annotations created inside WebViewer, their `Subject` have already been translated using i18n. See the code at `setupI18n.js`.
```js
window.Annotations.Utilities.setAnnotationSubjectHandler(type =>
  t(`annotation.${type}`),
);
```
Since it's already translated I feel like it's not necessary to do a `lowerCase` then `replace` here and then render `t(annotation.${val})`. `annotation.Subject` here is already `t(annotation.${val})`.

For annotations created by users in other PDF viewers, we don't have existing translations for their `Subject`s, so we don't need to do these string manipulations.

src/components/FilterAnnotMondal/FilterAnnotModal.js
```diff
@@ -0,0 +1,172 @@
...
+                      {[...annotTypes].map((val, index) => {
+                        return (
+                          <Choice
+                            type="checkbox"
+                            key={index}
+                            label={
+                              t(`annotation.${val}`)
+                                .toLowerCase()
+                                .replace(/\s/g, '') === val
+                                ? t(`annotation.${val}`)
```
ZhijieZhang, line 136: See my another comment regarding if we can just display `annot.Subject` here instead of getting the `val`s and use `t(annotation.${val})`

andreysaf:
image

Fixed CSS, fixed translations, removed closing of the modal on clear, fixed floating of the sorting container, fixed include author vs displayAuthor.

Thanks @ZhijieZhang for taking your time to review!

PR #824, opened 2020-08-06
[Feature] Added Math Symbol Pickers to Free text editor

zsydyc:
### Description

Add math symbol support to free text editor, see image below:
image

### Potential Risk
In my implementation, I used `editorRef.current.editor.setText()` to insert math symbol
Is this function going to be removed in product build? @ZhijieZhang ?

### Other questions:
Do I need to merge this to branch 7.0 ?


src/components/MathSymbolsPicker/MathSymbolsPicker.js
```diff
@@ -0,0 +1,56 @@
...
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+
+import selectors from 'selectors';
+
+import './MathSymbolsPicker.scss';
+
+const dataElement = 'colorPalette';
+
+class MathSymbolsPicker extends React.PureComponent {
```
jniemin, line 11: New components should use React function component rather than class components

src/components/MathSymbolsPicker/MathSymbolsPicker.js
```diff
@@ -0,0 +1,56 @@
...
+            {symbol}
+          </button>
+        ))}
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = state => ({
+  overridePalette: selectors.getCustomElementOverrides(state, dataElement),
```
ZhijieZhang, line 53: It looks like this prop is not used in this component. So maybe we can delete it?

src/components/MathSymbolsPicker/MathSymbolsPicker.js
```diff
@@ -0,0 +1,56 @@
...
+
+  render() {
+    const {
+      onClickHandler,
+    } = this.props;
+
+
+    const symbols = [
+      "\u002B", "\u2212", "\u00D7", "\u00F7", "\u003D", "\u2260", "\u00B1", "\u00AC", "\u003C", "\u003E", "\u22DC", "\u22DD", "\u00B0", "\u00B9", "\u00B2", "\u00B3", "\u0192", "\u0025", "\u0089", "\u2031", "\u2200", "\u2201", "\u2202", "\u2203", "\u2204", "\u2205", "\u2206", "\u2207", "\u2208", "\u2209", "\u220A", "\u220B", "\u220C", "\u220D", "\u220E", "\u220F", "\u2210", "\u2211", "\u2213", "\u2214", "\u2215", "\u2216", "\u2217", "\u2218", "\u2219", "\u221A", "\u221B", "\u221C", "\u221D", "\u221E", "\u221F", "\u2220", "\u2221", "\u2222", "\u2223", "\u2224", "\u2225", "\u2226", "\u2227", "\u2228", "\u2229", "\u222A", "\u222B", "\u222C", "\u222D", "\u222E", "\u222F", "\u2230", "\u2231", "\u2232", "\u2233", "\u2234", "\u2235", "\u2236", "\u2237", "\u2238", "\u2239", "\u223A", "\u223B", "\u223C", "\u223D", "\u223E", "\u223F", "\u2240", "\u2241", "\u2242", "\u2243", "\u2244", "\u2245", "\u2246", "\u2247", "\u2248", "\u2249", "\u224A", "\u224B", "\u224C"
+    ];
```
ZhijieZhang, line 28: nit: have line breaks for this array

src/components/MathSymbolsPicker/MathSymbolsPicker.js
```diff
@@ -0,0 +1,56 @@
...
+
+
+    return (
+      <div
+        data-element="MathSymbolsPicker"
```
ZhijieZhang, line 33: If a DOM element has the `data-element` attribute then it should be removed when clients call
```js
instance.disableElements(['MathSymbolPicker'])
```

We should either make it work with `disableElements`, or remove the `data-element` attribute.

assets/icons/preview.html
```diff
@@ -0,0 +1,687 @@
+<html>
```
ZhijieZhang, line 1: Can you show me where we used this file? I did a global search but didn't find... Might miss something.

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -192,6 +209,12 @@ const RichTextPopup = () => {
             img="ic_annotation_strikeout_black_24px"
             title="option.richText.strikeout"
           />
+          <Button
+            dataElement="richTextShowMathSymbols"
+            onClick={handleClick('showSymbols')}
```
ZhijieZhang, line 214: It looks like in `handleClick` we now have a special check for `showSymbols`. I prefer having a separate function for this and rename the original `handleClick` to something like `handleFormatChange` to remove the special check.

This is not a blocker but more like a nitpick, so no need to make a change if my suggestion isn't liked.

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -135,6 +141,17 @@ const RichTextPopup = () => {
   const syncDraggablePosition = (e, { x, y }) => {
     setDraggablePosition({ x, y });
   };
+
+
+  const insertSymbols = symbol => {
+    const contents = editorRef.current.getContents();
+    const { index, length } = editorRef.current.editor.getSelection();
```
ZhijieZhang, line 148: `current.editor.getSelection` will be gone in the production build. I think we should use `current.getSelection` instead.

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -135,6 +141,17 @@ const RichTextPopup = () => {
...
     setDraggablePosition({ x, y });
   };
+
+
+  const insertSymbols = symbol => {
+    const contents = editorRef.current.getContents();
+    const { index, length } = editorRef.current.editor.getSelection();
+    // only update the selected word (defined by index)
+    editorRef.current.editor.setText(`${contents.slice(0, index)}${symbol}${contents.slice(index + length)}`);
+    editorRef.current.editor.setSelection(index+1);
```
ZhijieZhang, line 151: `setText` and `setSelection` will be gone in the production build as they are not "exposed". More details can be found at `RichTextEditor.ts`.

For `setSelection` we can do `current.setSelection` instead of `current.editor.setSelection`.
For `setText` we would want to expose a wrapper method in the `RichTextEditor.ts`. For more details please take a look at other exposed methods in the file.

ZhijieZhang:
> Is this function going to be removed in product build? @ZhijieZhang ?
In general I would encourage us to run `npm run build` and then `npm run start-build` to see if there're any obfuscation issues after we finish implementing a feature. In this case the function will be obfuscated, so on the UI side it's "gone".

ZhijieZhang:
> Do I need to merge this to branch 7.0 ?

I wasn't aware that we are making this feature and don't know why we are doing this. So will let @mparizeau-pdftron to answer this question.

src/components/MathSymbolsPicker/MathSymbolsPicker.js
zsydyc: ok, I have converted this to a functional component.

src/components/MathSymbolsPicker/MathSymbolsPicker.js
zsydyc: This has been 

> It looks like this prop is not used in this component. So maybe we can delete it?

this prop is removed in my functional component refactor.

src/components/MathSymbolsPicker/MathSymbolsPicker.js
zsydyc: I removed the data-element attribute.

assets/icons/preview.html
zsydyc: We not really using this in any of our code, 
I have a hard time to preview all SVG icons, so I generated this preview.html using a Python script, and by looking at this preview.html, I can easily tell what does each SVG looks like.

I can also add the Python Script I used, if needed.

mparizeau-pdftron:
No we don't need it for 7.0

src/components/RichTextPopup/RichTextPopup.js
zsydyc: I agree. I rename the `handleClick` function to `handleTextFormatChange` and add a new function called `handleSymbolsClick`

src/components/RichTextPopup/RichTextPopup.js
zsydyc: Thanks for teaching.
I update Webviewer to expose a wrapper method `setText`, see PR:  https://github.com/XodoDocs/webviewer/pull/1599
Also I changed `current.editor.setSelection` to `current.setSelection` and `current.editor.setText` to `current.setText`

assets/icons/preview.html
ZhijieZhang, line 1: I see. Thanks for the clarification.

All the files under the `assets` folder are available to clients in the download package. It doesn't sound like the clients will need this preview file.

I suggest we either a) git ignore this file b) keep this file but add a comment at the beginning so internal developers and clients know the purpose of it.

src/components/MathSymbolsPicker/MathSymbolsPicker.js
```diff
@@ -0,0 +1,32 @@
...
+
+const symbols = [
+  "\u002B", "\u2212", "\u00D7", "\u00F7", "\u003D", "\u2260", "\u00B1", "\u00AC", "\u003C", "\u003E", "\u22DC", "\u22DD", "\u00B0", "\u00B9", "\u00B2", "\u00B3", "\u0192", "\u0025", "\u0089", "\u2031", "\u2200", "\u2201", "\u2202", "\u2203", "\u2204", "\u2205", "\u2206", "\u2207", "\u2208", "\u2209", "\u220A", "\u220B", "\u220C", "\u220D", "\u220E", "\u220F", "\u2210", "\u2211", "\u2213", "\u2214", "\u2215", "\u2216", "\u2217", "\u2218", "\u2219", "\u221A", "\u221B", "\u221C", "\u221D", "\u221E", "\u221F", "\u2220", "\u2221", "\u2222", "\u2223", "\u2224", "\u2225", "\u2226", "\u2227", "\u2228", "\u2229", "\u222A", "\u222B", "\u222C", "\u222D", "\u222E", "\u222F", "\u2230", "\u2231", "\u2232", "\u2233", "\u2234", "\u2235", "\u2236", "\u2237", "\u2238", "\u2239", "\u223A", "\u223B", "\u223C", "\u223D", "\u223E", "\u223F", "\u2240", "\u2241", "\u2242", "\u2243", "\u2244", "\u2245", "\u2246", "\u2247", "\u2248", "\u2249", "\u224A", "\u224B", "\u224C"
+];
+
+const MathSymbolsPicker = ({ onClickHandler }) => {
+
+  return (
+    <div
+      data-element="MathSymbolsPicker"
```
ZhijieZhang, line 14: https://github.com/PDFTron/webviewer-ui/pull/824#discussion_r466654982

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -1,29 +1,31 @@
...
+import Button from "components/Button";
+import HorizontalDivider from "components/HorizontalDivider";
+import { isMobile } from "helpers/device";
+import core from "core";
+import getRichTextPopupPosition from "helpers/getRichTextPopupPosition";
+import MathSymbolsPicker from "../MathSymbolsPicker";
+import actions from "actions";
+import selectors from "selectors";
+
+import "./RichTextPopup.scss";
```
ZhijieZhang, line 17: nit: string should use `''` instead of `""` for consistency

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -107,34 +117,49 @@ const RichTextPopup = () => {
...
 
   const syncDraggablePosition = (e, { x, y }) => {
     setDraggablePosition({ x, y });
   };
+
+  const insertSymbols = symbol => {
+    const contents = editorRef.current.getContents();
+    const { index, length } = editorRef.current.getSelection();
+    // only update the selected word (defined by index)
+    editorRef.current.setText(
```
ZhijieZhang, line 157: It looks like `setText` will remove all the formats. See the following gif:



On a user's perspective, I think the format of the symbol should be the format at the current selection index, and format of the rest of the text should be kept.

I haven't tried https://quilljs.com/docs/api/#inserttext but this API sounds like what we need?

assets/icons/preview.html
zsydyc: I will add this file to .gitignore.

src/components/RichTextPopup/RichTextPopup.js
zsydyc: I ran VS code file auto-format on this file, convert the double quote to single quote marks

src/components/RichTextPopup/RichTextPopup.js
zsydyc: This is fixed, use `deleteText` and `insertText` to replace `setText` function, please take a look and let me know.

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -40,27 +43,43 @@ const RichTextPopup = () => {
...
 
     core.addEventListener('editorSelectionChanged', handleSelectionChange);
-    return () => core.removeEventListener('editorSelectionChanged', handleSelectionChange);
+    return () =>
+      core.removeEventListener('editorSelectionChanged', handleSelectionChange);
   }, []);
 
   useEffect(() => {
     const handleTextChange = () => {
       if (annotationRef.current?.isAutoSized() && popupRef.current) {
```
ZhijieZhang, line 52: I think `spaceAdjustment` is always `0` since `symbolsVisible` is always `false` because `symbolsVisible` is missing in the dependency array.
However, When I was testing I didn't see anything weird with `spaceAdjustment` always being `0`. So I'm wondering if we actually need the adjustment?

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -40,27 +43,43 @@ const RichTextPopup = () => {
...
   }, []);
 
   useEffect(() => {
     const handleEditorFocus = (editor, annotation) => {
-      if (annotation instanceof window.Annotations.FreeTextAnnotation && popupRef.current) {
-        const position = getRichTextPopupPosition(annotation, popupRef);
+      if (
+        annotation instanceof window.Annotations.FreeTextAnnotation &&
+        popupRef.current
+      ) {
```
ZhijieZhang, line 75: Same comment here. `spaceAdjustment` is always 0

zsydyc:
> 2. I'm testing with Mac and see one empty symbol
>    Screen Shot 2020-08-10 at 3 27 53 PM
>    Not sure if this is intended?

The broken unicode are suppose to be "‰", but it doesn't seem to work on both Window and Mac. I replaced it will `\u2030` which represents the same character(‰) but displays well this time.

src/components/RichTextPopup/RichTextPopup.js
zsydyc: spaceAdjustment has been removed. I update `handleSymbolsClick` to adjust the popup position every time user show/hide the math symbols.  I feel the app is more smooth in this way.

assets/icons/preview.html
zsydyc: this file has been removed.

src/components/RichTextPopup/RichTextPopup.js
zsydyc: `spaceAdjustment` has been removed.  See my next commnet

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -107,13 +122,26 @@ const RichTextPopup = () => {
     return format;
   };
 
-  const handleClick = format => () => {
+  const handleTextFormatChange = format => () => {
     const { index, length } = editorRef.current.getSelection();
     const currentFormat = editorRef.current.getFormat(index, length);
 
     applyFormat(format, !currentFormat[format]);
   };
```
ZhijieZhang, line 131: Hmm didn't quite get why we need a `setTimeout` here. Can you explain a bit, and maybe add a comment in the code?

It'd be nice if we can get rid of it, as at a first glance this seems kind of unreliable.

If we need this `setTimeout` to re-position the popup so that the math symbol table doesn't get cut off, then I think a better way is to have another effect. It will look something like this:

```js
useEffect(() => {
 // calculate position
 // set css position
}, [symbolsVisible]);
```

The effect should run after the DOM has been updated. So when it runs we should have the correct height of the popup.

zsydyc:
Replace setTimeout by useEffect hook

src/components/RichTextPopup/RichTextPopup.js
```diff
@@ -107,13 +122,26 @@ const RichTextPopup = () => {
...
+  const handleTextFormatChange = format => () => {
     const { index, length } = editorRef.current.getSelection();
     const currentFormat = editorRef.current.getFormat(index, length);
 
     applyFormat(format, !currentFormat[format]);
   };
 
+  const handleSymbolsClick = () => {
+    setSymbolsVisible(!symbolsVisible);
+    setTimeout(() => {
```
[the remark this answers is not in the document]
zsydyc: This has been resolved by replacing `setTimeout` using `useEffect` hook

zsydyc:
This has been waiting for a while, can we merge both:
https://github.com/XodoDocs/webviewer/pull/1610
and 
this then this PR?

@mparizeau-pdftron  @jniemin

jniemin:
I think it good to merge

ZhijieZhang:
Yeah we can have an `enableFeatures` for this.

Examples can be found at `feature.js`, and `createFeatureAPI.js`.

zsydyc:
@ZhijieZhang I hide math-symbols feature by default and add `[Feature.MathSymbols]` to enable it, please take a look and let me know.

src/redux/actions/internalActions.js
```diff
@@ -242,6 +242,10 @@ export const setMouseWheelZoom = (enableMouseWheelZoom = true) => ({
   type: 'SET_MOUSE_WHEEL_ZOOM',
   payload: { enableMouseWheelZoom },
 });
+export const setMathSymbols = (enableMathSymbols = true) => ({
```
ZhijieZhang, line 245: As discussed in the screenshare before we actually don't need to add an action, reducer, and a selector for toggling the math symbols button. Since the code isn't necessary I think we should remove them.

zsydyc: I removed the changes from action and reducer, but I guess we still need the update inside `src/redux/selectors/exposedSelectors.js`, can you please take another look? @ZhijieZhang

zsydyc:
> Need to resolve conflicts.

merge conflict resolved, there will be some extra conflict after we merge: https://github.com/PDFTron/webviewer-ui/pull/804

ZhijieZhang:

PR #833, opened 2020-08-07
[Feature] - Allow the user to change the print quality in the print modal

zsydyc:
1. Add a feature to allow the user to change the print quality in the print modal
2. Also, run `npm run translation` to update strings

### DIFF
--- a/i18n/translation-zh_cn.json
+++ b/i18n/translation-zh_cn.json
@@ -207,6 +207,8 @@
       "current": "当前",
       "pages": "页数",
       "pageQuality": "打印质量",
+      "qualityNormal": "正常",
+      "qualityHigh": "高",
       "includeAnnotations": "包括注释",
       "includeComments": "包括评论",
       "addWatermarkSettings": "加水印"
--- a/i18n/translation-zh_cn.json
+++ b/i18n/translation-zh_cn.json
@@ -206,36 +206,37 @@
       "all": "全部",
       "current": "当前",
       "pages": "页数",
+      "pageQuality": "打印质量",
       "includeAnnotations": "包括注释",
       "includeComments": "包括评论",
-      "addWatermarkSettings": "添加水印设置"
+      "addWatermarkSettings": "加水印"
     },
     "redaction": {
-      "markForRedaction": "标记为编辑"
+      "markForRedaction": "标记为修订"
     },
     "searchPanel": {
       "caseSensitive": "区分大小写",
-      "wholeWordOnly": "整个词语",
+      "wholeWordOnly": "全字",
       "wildcard": "通配符"
     },
     "stampOverlay": {
       "addStamp": "添加新邮票"
     },
     "signatureOverlay": {
-      "addSignature": "添加签名"
+      "addSignature": "添加新签名"
     },
     "signatureModal": {
       "saveSignature": "保存签名",
       "dragAndDrop": "将图片拖放到此处",
       "or": "要么",
-      "pickImage": "选择签名图像"
+      "pickImage": "选择签名"
     },
     "status": {
       "status": "状态"
     },
     "state": {
       "accepted": "公认",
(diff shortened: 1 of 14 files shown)


src/components/PrintModal/PrintModal.js
```diff
@@ -18,6 +18,7 @@ import { Swipeable } from 'react-swipeable';
 import './PrintModal.scss';
 import Choice from '../Choice/Choice';
 import { FocusTrap } from '@pdftron/webviewer-react-toolkit';
+import setPrintQuality from 'src/apis/setPrintQuality';
```
andreysaf, line 21: Instead of APIs, we can dispatch an action to set the print quality. 

`store.dispatch(actions.setPrintQuality(quality));`

zsydyc:
> Great PR! Just a CSS comment and one code change.
> 
> Could we make the styling consistent and apply width:100% and height:30px. Also, please associate the label with the dropdown.
> 
> Before
> image
> 
> After
> image

This has been updated

i18n/translation-zh_cn.json
```diff
@@ -387,12 +390,12 @@
       "6": "星期六"
     },
     "weekdaysShort": {
-      "0": "周日",
-      "1": "周一",
-      "2": "周二",
-      "3": "周三",
+      "0": "太阳",
```
ZhijieZhang, line 393: Not sure what happened but the translations are kind of messed up now...

When you have time can you try regenerating them again? One way to do this is to copy-paste all the translation jsons in master into this branch, add the new `quality` translations, then run the script.

src/components/PrintModal/PrintModal.js
zsydyc: after discussed with Andrey, I replace this import function call to `onChange={e => store.dispatch(actions.setPrintQuality(e.target.value))}`

i18n/translation-zh_cn.json
zsydyc: Here is what I did:
1. Copied all `translation-xxx.json` file from current master branch
2. Add 3 `quality` strings to `translation-en.json` 
3. `npm run translation` to update all  `translation-xxx.json` file

After the translation strings list be build, I can see there are many updates to the existing strings, I did some check to the Chinese strings, most of the translations seem to make sense.

src/components/PrintModal/PrintModal.js
```diff
@@ -293,11 +305,24 @@ class PrintModal extends React.PureComponent {
...
+                        }))
+                      }
+                      checked={includeAnnotations}
                       center
                     />
                   </form>
+                  <div>
+                    <div className="col">
+                      <div>{`${t('option.print.pageQuality')}:`}</div>
+                      <select className="printQualitySelect" onChange={e => this.props.setPrintQuality(Number(e.target.value))}>
```
ZhijieZhang, line 320: Sorry one more thing that I noticed is that we probably want to set the `value` to be `this.props.printQuality`.

The reason is that clients are able to call `instance.setPrintQuality(x)` to set the print quality programmatically. In this case when they pass `2`, we want the modal to display `High`, not `Normal`.

If we set the `value` then we need to handle the case where clients pass in a number that's bigger than `2` to `instance.setPrintQuality`. In this case we want to display `High`, not `Normal`. I found one solution that works in Chrome is to move `<option value="2"` above `<option value="1"`... Not sure if it's a general solution.

zsydyc: I see your concern.
we only want to support printQuality between 1~2 from the UI, so if user set any value programmatically larger than 2, then we should always display "high'.
Move the `<option value="2"` will still display in this case. (Tested in Chrome, FF, Edge on Windows)

zsydyc:
@ZhijieZhang Just want to double-check if there is anything you want me to change before merge this in.

PR #839, opened 2020-08-12
[Bugfix] Annot subject is already translated, no need to translate it again

andreysaf:


andreysaf:
@ZhijieZhang since you caught it in my other PR, quick review for this bugfix.

Andrey

ZhijieZhang:
👍👍

PR #842, opened 2020-08-17
Added new API to set date format of printed notes

S0ulDrag0n:
Added a new API `setPrintedNoteDateFormat` to set the format of dates displayed on the notes page when printing.


ZhijieZhang:
LGTM

PR #845, opened 2020-08-18
[Bugfix] Order and show notes by create date

awy3:
This PR update the notes to sort by created date instead of "last activity" date when sorting by "time". This fix an issue with date shown not matching the header and annotation jumping to another section when updated, like the following

ModfiyingNote

This match what Acrobat is doing
image


ZhijieZhang:
@awy3 

If I'm reading correctly this PR fixes two things:
1. date shown not matching the header
This sounds like a bug to me. I'm assuming there's a bug in `src/helpers/getLatestActivityDate.js` that causes this issue. However I'm not clear why we chose to delete `getLatestActivityDate.js`and then implement a new `Time` sort logic instead of fixing the issue in that file?

2. annotation jumping to another section when updated
I checked the behavior in Acrobat, but I think it's tough to say that not matching Acrobat in this case is a bug in webviewer ui. Personally I think adding a comment to an annotation is updating it, so it makes sense that the root annotation is shown under `Today`.

PR #848, opened 2020-08-19
[Feature] Annotation line connector

bollain:
This PR implements the following:
https://trello.com/c/RN3RGaDD/960-when-annotation-is-selected-display-a-line-connecting-the-note-in-the-notes-panel-with-the-annotation-on-the-page

It has been requested by multiple clients as it was a feature available in 5.0/Jquery UI

Implemented in React using Portals.

Sample:
image

Now when the notes panel is open, and we select the note for an annotation, a line is drawn connecting the note with the annotation it is referring to. Selecting the annotation while the panel is open will also draw a line to connect its relevant note in the note panel.

Scrolling up and down makes the line disappear; it will only reappear if we either select the note or annotation.

The new element can be disabled/enabled using the relevant APIs; the new data element is `annotationLineConnector `

There are two cases where it may be useful to get some ideas from the team:
1) The line is made up of three segments, which allows the note to be connected on the annotation anywhere on the page, without changing the current behaviour of the note panel. When there is a very small y-coordinate difference between the note and the element, the vertical line is really small, making the overall connector look weird. What we could do here is have some logic to have the line be a single segments instead of three segments. Or we could leave it as is. 
Example:
image

2) When the zoom is at 200% + the segments start rendering weird, since there seems to be an issue with scaling of the widths of the annotations. In this instance we could disable the line; as it is probably not needed for this level of zoom. Or we could build in some logic to address these differences. 

While I could've made the call here, I think it's good to get some feedback and maybe have some of the team play around with this.


src/components/Note/Note.js
```diff
@@ -85,6 +88,8 @@ const Note = ({ annotation }) => {
       core.deselectAllAnnotations();
       core.selectAnnotation(annotation);
       core.jumpToAnnotation(annotation);
+      // Need this delay to ensure all other event listeners fire before we open the line
+      setTimeout(() => dispatch(actions.openElement('annotationLineConnector'), 300));
```
bollain, line 92: This was interesting. The timing of the events was causing the line to never open, as the `core.jumpToAnnotation` would trigger the onscroll event (effectively closing the line), but it was the last call, so after line 92 executed.

src/components/AnnotationLineConnector/AnnotationLineConnector.js
```diff
@@ -0,0 +1,93 @@
...
+import selectors from 'selectors';
+import { createPortal } from 'react-dom';
+
+import './AnnotationLineConnector.scss';
+
+
+const LineConnectorPortal = ({ children }) => {
+  const mount = document.getElementById("line-connector-root");
+  const el = document.createElement("div");
+  el.setAttribute('data-element', 'annotationLineConnector');
```
ZhijieZhang, line 14: ah I was wondering why this component can be disabled without specifically hook into the `useSelector(state => isElementDisabled(...))` logic. Turns out if an element is disabled then `openElement` won't do anything, then this component won't be rendered because of the `isLineOpen` check.

I'm not sure if we should still hook into the isDisabled logic to make things more explicit, even though it's not necessary.

src/components/AnnotationLineConnector/AnnotationLineConnector.js
```diff
@@ -0,0 +1,93 @@
...
+  const displayMode = core.getDisplayModeObject();
+  const annotationWindowCoords  = displayMode.pageToWindow({ x: annotation.getX(), y: annotation.getY() }, annotation.PageNumber);
+
+  useEffect(() => {
+    const { x: annotOriginX, y: annotOriginY } = annotationWindowCoords;
+    const { scrollTop } = core.getScrollViewElement();
+    const notePanelLeftPadding = 16;
+
+    // Annot Width to pixels
+    const annotationRightEdgeX = annotation.getX() + annotation.getWidth();
```
ZhijieZhang, line 55: If you want I think you could take advantage of the `getAnnotationPosition` function `getPopupPosition.js` to clean up the code here a bit(just for removing duplicates). I think that function calculates the rect of an annotation in window coordinates.

src/components/AnnotationLineConnector/AnnotationLineConnector.js
```diff
@@ -0,0 +1,93 @@
...
+
+    // Annot Height to pixels
+    const annotationBottomEdgeY = annotation.getY() + annotation.getHeight();
+    const annotBottomEdgeCoords = displayMode.pageToWindow({ x: annotation.getX(), y: annotationBottomEdgeY }, annotation.PageNumber);
+    const annotHeightInPixels = annotBottomEdgeCoords.y - annotOriginY;
+
+    setHorizontalLineOneRight(notePanelWidth - notePanelLeftPadding);
+    setHorizontalLineOneTop(noteContainerRef.current.getBoundingClientRect().top);
+
+    const lineWidth = window.innerWidth - notePanelWidth - annotOriginX + notePanelLeftPadding;
```
ZhijieZhang, line 67: What was the weird thing when the zoom level is bigger than 200?

From some testing I found that when the scroll view element starts to have horizontal scrolls, then there can be a gap at the right side of the annotation. To remove that gap we can take `scrollLeft` into account here.

```suggestion
    // get scrollLeft at line 51.
    const lineWidth = window.innerWidth - notePanelWidth - annotOriginX + notePanelLeftPadding + scrollLeft;
```

src/components/AnnotationLineConnector/AnnotationLineConnector.scss
```diff
@@ -0,0 +1,16 @@
...
+background-color: rgba(30,120,235, 0.5);;
+position: fixed;
+height: 2px;
+}
+
+.lineVertical {
+background-color: rgba(30,120,235, 0.5);;
+position: fixed;
+width: 2px;
+}
```
ZhijieZhang, line 16: Format is a bit off here.
```suggestion
#line-connector-root {
  position: relative;
  z-index: 999;
}

.lineHorizontal {
  background-color: rgba(30, 120, 235, 0.5);
  position: fixed;
  height: 2px;
}

.lineVertical {
  background-color: rgba(30, 120, 235, 0.5);
  position: fixed;
  width: 2px;
}

```

src/components/App/App.js
```diff
@@ -86,7 +86,7 @@ const App = ({ removeEventHandlers }) => {
 
         <Header />
         <ToolsHeader />
-        <div className="content">
+        <div className="content" onScroll={() => dispatch(actions.closeElement('annotationLineConnector'))}>
```
ZhijieZhang, line 89: Another place to put this line is

https://github.com/PDFTron/webviewer-ui/blob/2c7cfd4461285f9e1d7ef4931a72fd8c7294d62c/src/components/DocumentContainer/DocumentContainer.js#L181-L186

Might be better to put this line there for colocation.

bollain:
@ZhijieZhang thanks for the review! Made changes based on your suggestions.

I figured out the zoom issue - I wasn't subtracting the annotation width from the complete line (just one segment), which was causing issues at really high zoom levels for non-note annotations. This is now fixed!

ZhijieZhang:

PR #849, opened 2020-08-19
bugfix/add page-break-inside: avoid to .note CSS class

CorreyL:
## Preamble

I made this change for a client who needed the comments not to be rendered across a page break, and this is the solution I suggested as a start.

image

Left is before, right is after.

Haven't had time to test the greater implications of `display: flex` to `display: table` (`page-break-inside: avoid` doesn't play nicely with `display: flex`), so I figured I'd put this up as a Draft PR, and if it becomes an issue that needs to be fixed in the future, someone on the WebViewer team could take a closer look and see if there's more robust solution.

### DIFF
--- a/src/components/PrintHandler/PrintHandler.scss
+++ b/src/components/PrintHandler/PrintHandler.scss
@@ -44,7 +44,7 @@
 
     .page__container {
       box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


ZhijieZhang:

PR #865, opened 2020-08-25
[CHANGE] fixed issue where loadDocument(...., {streaming: true}) was …

achenPDFTron:
…not streaming correctly

for https://support.pdftron.com/a/tickets/15918

### DIFF
--- a/src/helpers/loadDocument.js
+++ b/src/helpers/loadDocument.js
@@ -33,7 +33,7 @@ const getDefaultOptions = () => ({
   forceClientSideInit: getHashParams('forceClientSideInit', false),
   disableWebsockets: getHashParams('disableWebsockets', false),
   cacheKey: JSON.parse(getHashParams('cacheKey', null)),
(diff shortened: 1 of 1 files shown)


src/helpers/loadDocument.js
```diff
@@ -80,7 +80,7 @@ const extractXodOptions = options => {
   }
 
   if (options.streaming !== null) {
-    xodOptions.streaming = options.streaming === 'true';
+    xodOptions.streaming = options.streaming;
```
achenPDFTron, line 83: it looks like if we have streaming set on constructor, the value is a string and then it gets set here.

before we had 
xodOptions.streaming = options.streaming === 'true';

so if we call loadDocument with streaming set to true,

then options.streaming === 'true' becomes false.

ZhijieZhang:
Closing because PRs should be opened in the webviewer repo instead. Let me know if you have any questions!

PR #884, opened 2020-10-14
[Bugfix] Fix text overflow for author names

awy3:
THIS PR ISN'T INTENDED TO BE COMMITTED, this is for showing someone how to fix this issue in their custom build. Another PR will fix this issue


ZhijieZhang:
@awy3 Do we still need this PR?

PR #892, opened 2021-01-11
get print quality from store for printInBackground too

mucahiddayan:


mbaer3000:
I suggest we keep the change as small as possible, no prettier edits etc. That also makes reviewing much easier.

mucahiddayan:
will remove prettier edits

PR #894, opened 2021-01-12
fix issue related to printQuality in printInBackground

mucahiddayan:


src/helpers/print.js
```diff
@@ -94,11 +94,11 @@ const printPdf = () =>
       });
   });
 
-export const creatingPages = (pagesToPrint, includeComments, includeAnnot, printQualty, sortStrategy, clrMap, dateFormat, onProgress) => {
+export const creatingPages = (pagesToPrint, includeComments, includeAnnot, printQuality, sortStrategy, clrMap, dateFormat, onProgress) => {
   const createdPages = [];
   pendingCanvases = [];
   includeAnnotations = includeAnnot;
-  printQuality = printQualty;
+  PRINT_QUALITY = printQuality;
```
mbaer3000, line 103: Would it suffice to drop this line and simply have `multiplier: printQuality` in line 189?

src/helpers/print.js
```diff
@@ -50,7 +50,7 @@ export const print = async(dispatch, isEmbedPrintSupported, sortStrategy, colorM
       pagesToPrint,
       includeComments,
       includeAnnotations,
-      printQuality,
+      printQuality||PRINT_QUALITY,
```
mbaer3000, line 54: Better have default assignment in line 20, as in `({ includeAnnotations, includeComments, pagesToPrint, onProgress, printWithoutModal, printQuality = PRINT_QUALITY} = options)`

src/helpers/print.js
```diff
@@ -10,14 +10,14 @@ import core from 'core';
...
+let PRINT_QUALITY = 1;
 let colorMap;
 
 export const print = async(dispatch, isEmbedPrintSupported, sortStrategy, colorMap, options) => {
-  let includeAnnotations, includeComments, pagesToPrint, onProgress;
+  let includeAnnotations, includeComments, pagesToPrint, onProgress, printQuality;
   let printWithoutModal = false;
   if (options) {
-    ({ includeAnnotations, includeComments, pagesToPrint, onProgress, printWithoutModal } = options);
+    ({ includeAnnotations, includeComments, pagesToPrint, onProgress, printWithoutModal, printQuality } = options);
```
mbaer3000, line 21: Drop if condition and rather have default assignment in method signature, as in `async(dispatch, isEmbedPrintSupported, sortStrategy, colorMap, options = {}) => {…}`

src/helpers/print.js
```diff
@@ -10,14 +10,14 @@ import core from 'core';
 
 let pendingCanvases = [];
 let includeAnnotations = false;
-let printQuality = 1;
+let PRINT_QUALITY = 1;
```
mbaer3000, line 13: We should name this `DEFAULT_ PRINT_QUALITY`

mucahiddayan:
newer PR exists

PR #908, opened 2021-03-03
Updated Dutch translation into a more appropriate one.

martis347:
Hello!
We're using PDFTron in our application and some of our users are Dutch. Today we received a support ticket asking to update a translation that's not entirely correct, so I'm opening a Pull Request to update that.

From Google Translate:
"Geen een wachtwoord" translates to "Not a password".
"Geef een wachtwoord" translates to "Enter a password".


ZhijieZhang:
Sorry for the delay and thank you very much for opening this PR.

PR #913, opened 2021-04-22
Fixed inconsistent en translation for noAnnotations

Kuirak:


khein-pdftron:
Hello @Kuirak . Thank you for pointing this out. The english text is actually correct and we forgot to update the other languages. I will close this and we will make those updates.

Kuirak:
If you do this please add a different Text for readonly Mode. Since the Text is wrong in read only context.

khein-pdftron:
That is a good point. We will use your text for read only mode.

PR #916, opened 2021-05-20
Use Dart Sass instead of Node Sass

naueramant:
Use Dart Sass (sass) instead of Node Sass (node-sass) as recommended by node sass https://www.npmjs.com/package/node-sass

> Warning: LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

Also the sass compiler has been moved to DevDependencies as it should not be necessery as a Dependency.


mparizeau-pdftron:
Thanks! We're going to do a little testing but it looks like a good change and we should be able to merge it in soon.

bollain (approved):
Merging into master branch - changes will be part of 8.0 release

PR #953, opened 2022-01-25
added action to select page form, so iOS handles it correctly

radekhvitrium:
This is needed, with the action attribute the keyboard shows "Go" instead of "Return" on iOS (without it, it does not have to call onSubmit event.)


bollain:
Hi @radekhvitrium, 

Thanks for this PR! I did a bit of research and it seems that having an empty action attribute is not advised. However, I did find that we can simply add the enterkeyhint property to the input to have control over what shows up in the mobile keyboard:
https://stackoverflow.com/a/39485162/12962928

Based on this, we prefer to have the enterkeyhint property. I will add it from my end and will close this PR. Once it is merged I can drop a note here.

bollain:
@radekhvitrium update: I have added this to our backlog to add these attributes to all the inputs in our UI where it makes sense. If you wish to have this one updated right away, please submit a PR.

PR #954, opened 2022-02-10
Update translation-nl.json

stayallive:
I went through the Dutch translations (as a native Dutch speaker) to update the most notable translation artifacts where words were translated a little too literally and now make more sense in their context.

If this should be done elsewhere please let me know where I can submit these language suggestions.


bollain:
Thanks for this PR @stayallive! I will merge them now. They will be available from our nightly stable builds in about a day or so. They will also be part of our upcoming 8.3 release.

bollain (approved):

PR #955, opened 2022-02-23
print page limit feature

radekhvitrium:
I have implemented print page limit feature to our fork. I am creating a pull request so if you are interested, you can merge this in your repository.

The problem with the print was that if you have a large 5000 pages document and you are trying to render all pages for print, the resources of a browser are not enough and the print fails. Now you can do this: 
`viewer.UI.setPrintPageLimit(100);`
which will render only 100 pages, then the browser print dialog is shown and after closing it it will render another 100 pages another browser print dialog and so on.


radekhvitrium:
Does not work on mobile, if I will resolve this, I will create another pull request

PR #982, opened 2022-06-20
Update translation-ko.json

JoYongpil:
update korean translaion

### DIFF
--- a/i18n/translation-ko.json
+++ b/i18n/translation-ko.json
@@ -1,6 +1,6 @@
 {
   "action": {
-    "add": "추가하다",
+    "add": "추가",
     "apply": "적용",
     "applyAll": "모두적용",
     "calendar": "캘린더",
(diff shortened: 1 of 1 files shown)


i18n/translation-ko.json
```diff
@@ -289,7 +289,7 @@
       "toolbarGroup-Annotate": "주석",
       "toolbarGroup-Shapes": "도형",
       "toolbarGroup-Insert": "삽입",
-      "toolbarGroup-Measure": "측정",
+      "toolbarGroup-Measure": "단위",
```
bollain, line 292: Based on its context we think it should be:
```suggestion
      "toolbarGroup-Measure”: “측정",
```

image

i18n/translation-ko.json
```diff
@@ -799,7 +799,7 @@
     "cropDimensions": "크기",
     "dimensionInput": {
       "unitOfMeasurement": "단위",
-      "autoTrim": "“자동"
+      "autoTrim": "자동트림"
```
bollain, line 802: Based on context we have the following suggestion:
```
“autoTrim”: “자동트림” => “자동 자르기”  or “자동”
```

image

bollain (changes requested):
Hi @JoYongpil, thanks again for this PR. We have a couple of suggestions based on an internal review, but otherwise this looks good.

bollain (approved):
Thanks for submitting this @JoYongpil. If you are building the UI locally, these will be available immediately. If you are using our hosted builds, they should be available by next week.

PR #993, opened 2022-07-18
8.7 update

awy3:
The 8.7 version update


awy3:
Wrong PR, shouldn't be merging this into 8.6. Will just be making tag of branch

PR #995, opened 2022-09-21
Manual copy from latest WebViewer 8.9

wanboli:


wanboli:
@mparizeau-pdftron Looks like we have lots of files un-synced, and also somehow the lint is failing now. Is there any configuration we should do here?
Thanks.

bollain (approved):

PR #1007, opened 2023-04-27
Manual copy from latest WebViewer 10.1

wanboli:


wanboli:
As talked to Armando, will sync the repo after v10.1 is released. Close the PR for now.

PR #1021, opened 2023-06-12
improvements to assigning type to redaction search results

kanderson91:
The webviewer assigns a type to each redaction search result by iterating through all the patterns and seeing which one matches the result text. I was running into some problems with this so I made the following changes:

-Keep track of which patterns are in the current search so that we only have to iterate through those.
-If the user is doing a search for just one pattern, always assign the type of that pattern to each search result.
-When checking if a pattern matches the search result, use the ambient text instead of just the result string. 

That last change is useful when using patterns with lookarounds in them. For example, say the user adds this custom pattern to match any string of digits preceded by "ID:"
 (?<=ID:\s)\d+

Then say the page has the text "ID: 123" and so we get a search result with "123". In the current logic, the system will check if  (?<=ID:\s)\d+ matches "123" and find that it does not. So I changed it to check if it matches the ambient text "ID: 123".

### DIFF
--- a/src/components/RedactionSearchPanel/RedactionSearchPanel.js
+++ b/src/components/RedactionSearchPanel/RedactionSearchPanel.js
@@ -1,10 +1,9 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext } from 'react';
 import RedactionSearchOverlay from 'src/components/RedactionSearchOverlay';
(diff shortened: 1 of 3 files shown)


bollain:
Hey @kanderson91, thanks for taking the time to make this PR! We are reviewing it internally and will let you know once this is completed.

PR #1056, opened 2024-04-30
Correct some german translations

mithyB:
Corrected some german translations to make grammatically sense where they are used on the UI


bollain:
Hi @mithyB,

Thanks for these suggestions! We are reviewing them internally, I'll update this PR once this is ready to be merged.

bollain:
Thanks again @mithyB - I will merge these. They will be also be part of our upcoming 10.10 release.

PR #1070, opened 2024-07-02
[Bugfix] instance.UI.setActiveLeftPanel not working in modular UI

### LINKED ISSUE #1069: [Bug] instance.UI.setActiveLeftPanel not working in modular UI
When calling `instance.UI.setActiveLeftPanel` in combination with the modular UI, I encountered some strange behaviour.

**WebViewer version**
- 10.10.1 both UI and Core
- Modular UI (ui set to "beta")

**The current behavior**
There are no changes in regard to the active panel, not just when calling it with "thumbnailsPanel" but any other (valid) panel of yours.

Also, the browser console shows the following warning:

image

**The expected behavior**
- No console warning
- the panel that is supposed to be set is _actually_ set to active

**Steps to reproduce**
Call `instance.UI.setActiveLeftPanel("thumbnailsPanel")` and take a look at the browser console

**Solution Idea**
I had a look at the code (`./src/helpers/isDataElementLeftPanel.js`) and noticed that it would probably only require the following change:

```js
  if (customizableUI) {
    return state.viewer.genericPanels
      .filter((item) => item.location === PLACEMENT.LEFT)
      .map((item) => item.dataElement);
  }
```

instead of

```js
  if (customizableUI) {
    return state.viewer.genericPanels.filter((item) => item.location === PLACEMENT.LEFT);
  }
```

as you filter for inclusion of concrete data-elements afterwards.

Olliwehr:
- Fixes #1069 

Concrete data-element identifier like "thumbnailsPanel" are obviously not included in a collection that consists of objects that have the form

```js
{
  dataElement ...
  location ...
  render: ...
}
```

So mapping these down to their dataElement property should make the inclusion check work again :)

I ran `npm run lint` before but because of just a single changed line I think there isn't much work to be done here 😄

### DIFF
--- a/src/helpers/isDataElementLeftPanel.js
+++ b/src/helpers/isDataElementLeftPanel.js
@@ -8,7 +8,7 @@ export const getLeftPanelDataElements = (state) => {
   const { customizableUI } = featureFlags;
 
   if (customizableUI) {
-    return state.viewer.genericPanels.filter((item) => item.location === PLACEMENT.LEFT);
+    return state.viewer.genericPanels.filter((item) => item.location === PLACEMENT.LEFT).map((item) => item.dataElement);
   }
 
   const defaultLeftPanels = [


Olliwehr:
Uhh, quickly switched back to 10.10 as base as switching to master showing some strange diffs. Hit me up if I should change the base/PR to your desired target! 🙂

Olliwehr:
It might need further testing, as I think there are some troubles with the left panel not showing up correctly after my change. :(
But I did not see any further errors, so don't know why this happens.

bollain:
Closing as we have a new API here:
https://github.com/ApryseSDK/webviewer-ui/blob/11.1/src/apis/setActiveTabInPanel.js

Thanks again for bringing this to our attention!

PR #1095, opened 2025-04-11
Update TouchEventManager.js

petryshkaCODE:
fix old link to a new one


bollain (commented):
Thanks for this update @petryshkaCODE. Do you happen to be the maintainer of that blog?

